PR #1, opened 2021-07-19
Adding docker file

reetika-vyas:


T0MASD:
Please take a look at https://github.com/openshift/ocm-container/blob/master/Dockerfile for inspiration; it uses 2 step build process and has section on installing oc

reetika-vyas:
> Please take a look at https://github.com/openshift/ocm-container/blob/master/Dockerfile for inspiration; it uses 2 step build process and has section on installing oc

Thanks, Tomas, will be making changes as suggested

PR #3, opened 2021-07-21
OSD 7660

reetika-vyas:
https://issues.redhat.com/browse/OSD-7660


backplane/project.mk
```diff
@@ -0,0 +1,3 @@
+IMAGE_REGISTRY?=quay.io
+IMAGE_REPOSITORY?=openshift-sre
```
wanghaoran1988, line 2: app-sre?

backplane/project.mk
```diff
@@ -0,0 +1,3 @@
...
+IMAGE_NAME?=backplane-m3
```
wanghaoran1988, line 3: managed-scripts

wanghaoran1988:
@qhua948 Could you please take look if you have time?

backplane/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Copy previously acquired binaries into the $PATH
+ENV BIN_DIR="/usr/local/bin"
+RUN mv /out/oc ${BIN_DIR}
+
+# Validate
+RUN oc completion bash > /etc/bash_completion.d/oc
+
+# Cleanup Home Dir
+RUN rm /root/anaconda* /root/original-ks.cfg
```
wanghaoran1988, line 41: what's this? where it's coming from?

backplane/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Make binaries executable
+RUN chmod -R +x /out
+
+# Copy previously acquired binaries into the $PATH
+ENV BIN_DIR="/usr/local/bin"
+RUN mv /out/oc ${BIN_DIR}
+
+# Validate
+RUN oc completion bash > /etc/bash_completion.d/oc
```
wanghaoran1988, line 38: Start from line 10, it's all about install "oc", could we write a simple script, and invoke that in the Dockerfile

wanghaoran1988:
Could you rename backplane -> hack?

backplane/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Pre-install packages
+FROM quay.io/centos/centos:latest
```
ravitri, line 2: We shouldn't use centos image. Should use any image which is under https://catalog.redhat.com/software/containers/explore and is supported by Redhat.

ravitri:
I have a query. Why not directly use existing image like registry.redhat.io/openshift4/ose-cli ? (Refer https://catalog.redhat.com/software/containers/openshift4/ose-cli/5cd9ba3f5a13467289f4d51d?container-tabs=gti). If eventually we need an image with `oc` then we already have this image. The only issue I see here is that the image is not part of quay repo so can't use existing token. If we can mirror this or clone the image to quay from time to time then that might work too.

backplane/build/Dockerfile
georgettica: I think this comes from a default rhel/fedora installation

hack/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Pre-install packages
+FROM quay.io/centos/centos:latest
+
+# Install packages
+RUN yum update -y \
+    yum upgrade -y \
```
qhua948, line 6: Run a `yum clean all` https://stackoverflow.com/questions/46089219/how-to-reduce-the-size-of-rhel-centos-fedora-docker-image

hack/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Download the checksum
+RUN curl -sSLf ${OC_URL}/sha256sum.txt -o sha256sum.txt
+
+# Download the binary tarball
+RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux" '$0~asset {print $2}' sha256sum.txt)"
+
+# Check the tarball and checksum match
+RUN sha256sum --check --ignore-missing sha256sum.txt
+RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
```
qhua948, line 28: I suggest using docker multi stage build, make things cleaner

hack/build/Dockerfile
qhua948: Can you also install python3 as well? We have intention to use that as a target backplane m3 language

hack/build/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# Copy previously acquired binaries into the $PATH
+ENV BIN_DIR="/usr/local/bin"
+RUN mv /out/oc ${BIN_DIR}
+
+# Validate
+RUN oc completion bash > /etc/bash_completion.d/oc
+
+# Cleanup Home Dir
+RUN rm /root/anaconda* /root/original-ks.cfg
+
```
qhua948, line 42: We will also need to copy the scripts files into the docker image somewhere
maybe under `/managed-scripts/` ?

qhua948 (changes requested):

hack/build/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
+
+# Make binaries executable
+RUN chmod -R +x /out
+
+FROM registry.access.redhat.com/ubi8/ubi:latest
+RUN  yum -y install --disableplugin=subscription-manager \
+  python3 \
+  && yum --disableplugin=subscription-manager clean all
+COPY --from=build-stage0 /out/oc  /out/oc
```
qhua948, line 29: Can just do a single step copy

hack/build/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ENV BIN_DIR="/usr/local/bin"
+RUN mv /out/oc ${BIN_DIR}
+
+# Validate
+RUN oc completion bash > /etc/bash_completion.d/oc
+
+# Cleanup Home Dir
+RUN rm /root/anaconda* /root/original-ks.cfg
+WORKDIR /root
+ENTRYPOINT [ "/bin/bash", "-l", "-c" ]
```
qhua948, line 41: why do we need a login shell here?

hack/build/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux" '$0~asset {print $2}' sha256sum.txt)"
+
+# Check the tarball and checksum match
+RUN sha256sum --check --ignore-missing sha256sum.txt
+RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
+
+# Make binaries executable
+RUN chmod -R +x /out
+
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
qhua948, line 25: I think we can use minimal? https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=overview

qhua948: More docs: https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image

qhua948 (approved):

PR #10, opened 2021-08-31
Jsonschema example

wanghaoran1988:


hack/metadata.schema.json
```diff
@@ -1,114 +1,735 @@
...
+                                                                            "pods"
+                                                                        ]
+                                                                    }
+                                                                ]
+                                                            }
+                                                        },
+                                                        "resourceNames": {
+                                                            "$id": "#/properties/rbac/properties/roles/items/anyOf/0/properties/rules/items/anyOf/0/properties/resourceNames",
+                                                            "type": "array",
+                                                            "title": "The resourceNames schema",
```
qhua948, line 428: Need to make this optional

wanghaoran1988: this schema is auto generated, and I made no changes to that, this is just for testing, @qhua948 Could you update the existing one that already in the repo?

qhua948: :+1:

hack/app_sre_pr_check.sh
```diff
@@ -1,7 +1,32 @@
...
+do
+   echo "convert $f to json"
+   if ! python3 hack/yamltojson.py $f; then
+     echo "convert failed"
+     exit 1
+   else
+     echo "convert succeed"
+   fi
+done
+#Verify the metadata.json are valide
```
qhua948, line 19: s/valide/valid

hack/app_sre_pr_check.sh
```diff
@@ -1,7 +1,32 @@
...
+jfiles=$(find . -name 'metadata.json')
+for f in $jfiles
+do
+   echo "validating the jsonschema for $f"
+   if ! $CONTAINER_ENGINE run --rm -v $(pwd):/json quay.io/haowang/jsonschema:latest -i /json/$f /json/hack/metadata.schema.json; then
```
qhua948, line 24: can we invoke a single container and then run a `find ... exec ...` in there? I think it's better than invoking docker each time

hack/app_sre_pr_check.sh
```diff
@@ -1,7 +1,32 @@
...
+     echo "validating failed: $f"
```
qhua948, line 25: s/validating/validation

hack/app_sre_pr_check.sh
wanghaoran1988: This is more easy for me ,we can fail at anytime when the validation failed.

PR #14, opened 2021-09-09
Update makefile

wanghaoran1988:


hack/app_sre_pr_check.sh
```diff
@@ -2,31 +2,4 @@
...
-   if ! $CONTAINER_ENGINE run --rm -v $(pwd):/json quay.io/haowang/jsonschema:latest -i /json/$f /json/hack/metadata.schema.json; then
-     echo "validation failed: $f"
-     exit 1
-   else
-     echo "validation succeed"
-   fi
-done
-
-make IMAGE_REPOSITORY=${IMAGE_REPOSITORY:-app-sre} build
+make IMAGE_REPOSITORY=${IMAGE_REPOSITORY:-app-sre} build validation
```
qhua948, line 5: no need validation here, since build depends on validation

hack/schema_validation.sh
```diff
@@ -0,0 +1,31 @@
...
+   else
+     echo "convert succeed"
+   fi
+done
+#Verify the metadata.json are valid
+jfiles=$(find . -name 'metadata.json')
+for f in $jfiles
+do
+   echo "validating the jsonschema for $f"
+   if ! $CONTAINER_ENGINE run --rm -v $(pwd):/json quay.io/haowang/jsonschema:latest -i /json/$f /json/hack/metadata.schema.json; then
```
feichashao, line 24: is `haowang` expected here?

wanghaoran1988: yes, it's in my own repo, we need move this somewhere finally.

hack/app_sre_pr_check.sh
wanghaoran1988: @qhua948 Done

PR #17, opened 2021-09-18
Fix warnings in etcd health check

feichashao:
Remove `--cluster` and `-it` in the script to avoid warnings.


feichashao:
wrong commits, closing this

PR #23, opened 2021-10-18
feat(olm-dance): add olm-dance

georgettica:
now you can run the same script that was in https://github.com/openshift/ops-sop/blob/master/v4/utils/reset-olm.sh

### DIFF
--- a/scripts/SREP/delete-olm-operator/metadata.yaml
+++ b/scripts/SREP/delete-olm-operator/metadata.yaml
@@ -1,0 +1,32 @@
+file: script.sh
+name: delete-olm-operator 
+description: Delete all of the olm data in a given namespace (if syncsets are working, it will recreate)
+author: gerogettica
+allowedGroups: 
+  - SREP
+  - CSSRE
+  - CEE
+  - MTSRE
+rbac:
+    roles: {}
+    clusterRoleRules:
+        - verbs:
+            - "get"
+            - "list"
+            - "watch"
+            - "delete"
+          apiGroups:
+            - "operators.coreos.com/v1alpha1"
+          resources:
+            - "clusterserviceversions"
+            - "catalogsources"
+            - "operatorgroups"
+            - "subscriptions"
+envs:
+  - key: "NAMESPACE"
+    description: "the namespace in which all of the operators will run"
+    optional: false
+  - key: "FORCE"
+    description: "without setting this, the script will just display what it was going to do"
+    optional: true
+language: bash
--- a/scripts/SREP/delete-olm-operator/script.sh
+++ b/scripts/SREP/delete-olm-operator/script.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+set -e
+set -o nounset
+set -o pipefail
+
+DELETE=false
+if [[ -n "${FORCE}" ]]; then
+	DELETE=true
+fi
+
+"$DELETE" || echo "Not going to delete resources"
+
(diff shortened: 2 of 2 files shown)


georgettica:
issue connecting to `mirror.openshift.com`
/retest

scripts/SREP/delete-olm-operator/script.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "Deleting OperatorGroup: ${OPERATOR_GROUP}"
+    oc delete og "${OPERATOR_GROUP}" -n "${NAMESPACE}" 
+  fi
+else
+  echo "Will delete Subscription: ${SUBSCRIPTION}"
+  echo "Will delete CatalogSource: ${CATALOG_SOURCE}"
+  echo "Will delete OperatorGroup: ${OPERATOR_GROUP}"
+fi
+
+$DELETE || echo "Will delete CSVs:"
```
qhua948, line 38: here as well

scripts/SREP/delete-olm-operator/metadata.yaml
```diff
@@ -0,0 +1,32 @@
+file: script.sh
+name: delete-olm-operator 
+description: Delete all of the olm data in a given namespace (if syncsets are working, it will recreate)
+author: gerogettica
+allowedGroups: 
+  - SREP
+  - CSSRE
+  - CEE
```
wanghaoran1988, line 8: Do we wanna CEE do this operation?

wanghaoran1988: We can only keep SREP here, as this used by SREP only atm.

scripts/SREP/delete-olm-operator/metadata.yaml
```diff
@@ -0,0 +1,32 @@
...
+            - "delete"
+          apiGroups:
+            - "operators.coreos.com/v1alpha1"
+          resources:
+            - "clusterserviceversions"
+            - "catalogsources"
+            - "operatorgroups"
+            - "subscriptions"
+envs:
+  - key: "NAMESPACE"
```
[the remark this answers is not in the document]
wanghaoran1988: by checking whether it's prefixed with "openshift-*" or "redhat-*" ?

scripts/SREP/delete-olm-operator/metadata.yaml
georgettica: Done

qhua948:
> What does that mean? What is the bash built-in for true and false?
> 
> -- meaning, how do I do the chage? @qhua948

So when you do a 
`"true" || cmd`, bash is looking for the `true` command on `$PATH`, this call external and not portable and should be avoided.

You should use the bash build in extended test `[[ cond ]]` instead

georgettica:
I tried to run the following command with PATH='' and it succeeded
```bash
#!/bin/bash
a=true
"$a" || echo yes
"$a" && echo no
```

georgettica:
Anyways found you have the builtin keyword that will run only builtin commands

qhua948:
> Anyways found you have the builtin keyword that will run only builtin commands

Thanks! 

I thought for sure it was calling the external cmd.

georgettica:
It's not, and it seems that the bash docs clearly state it starts by checking the built-ins first
https://www.gnu.org/software/bash/manual/bash.html#Command-Search-and-Execution

qhua948:
feel free to revert to the version without calling `builtin`

PR #29, opened 2021-11-02
Move etc health check script

qhua948:


wanghaoran1988:
@qhua948 Why we move to different dir?

PR #30, opened 2021-11-03
OSD-8555 allow CEE to read node-logs

feichashao:
Allow CEE to read node-logs.

The permission `nodes/proxy` is too wide, so we use a wrapper managed-script to allow them to read journal log on that node only.


georgettica:
Why not just add these permissions in MCC?

georgettica:
Bvm, answered myself by reading the title

georgettica:
/lgtm
/hold
Holding as it might be better for someone with that use case to test this

PR #31, opened 2021-11-07
[TODO] feat(daemonset-fix): initial script

georgettica:
related to alert I got, and it's still very raw, but if someones sees use it it let me know


georgettica:
Not getting to it, so I'll close (maybe this will come in handy for someone)

PR #32, opened 2021-11-07
[OSD-8686] Convert v4-registry-s3-size script to managed-script

karthikperu7:
https://issues.redhat.com/browse/OSD-8686

Also fixed the original script from ops-sop here: https://github.com/openshift/ops-sop/pull/1563

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,16 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS build-stage0
 ARG OC_VERSION="stable"
 ENV OC_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OC_VERSION}"
 
+# AWS cli args and env variables
+# Replace AWS client zipfile with specific file to pin to a specific version
+# (eg: "awscli-exe-linux-x86_64-2.0.30.zip")
+ARG AWSCLI_VERSION="awscli-exe-linux-x86_64.zip"
+ENV AWSCLI_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}"
+ENV AWSSIG_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}.sig"
+
+# install tools needed for installation
+RUN yum install -y unzip
+
 # Directory for the extracted binary
 RUN mkdir -p /out
 
@@ -19,6 +29,18 @@ RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux
 RUN sha256sum --check --ignore-missing sha256sum.txt
 RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
 
+# Install aws-cli
+RUN mkdir -p /aws/bin
+WORKDIR /aws
+# Download the awscli zip file
+RUN curl -sSLf $AWSCLI_URL -o awscliv2.zip
+# Extract the awscli zip
+RUN unzip awscliv2.zip
+# Install the libs to the usual location, so the simlinks will be right
+# The final image build will copy them later
+# Install the bins to the /aws/bin dir so the final image build copy is easier
+RUN ./aws/install -b /out
+
 # Make binaries executable
 RUN chmod -R +x /out
 
@@ -27,6 +49,7 @@ RUN  yum -y install --disableplugin=subscription-manager \
      python3  \
      && yum --disableplugin=subscription-manager clean all
 COPY --from=build-stage0 /out/oc  /usr/local/bin
+COPY --from=build-stage0 /out/aws  /usr/local/bin
 COPY scripts /managed-scripts
 
 # Validate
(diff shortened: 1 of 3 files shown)


karthikperu7:
the script requires `aws` cli so installing that in the Dockerfile

Dockerfile
```diff
@@ -2,6 +2,16 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS build-stage0
 ARG OC_VERSION="stable"
 ENV OC_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OC_VERSION}"
 
+# AWS cli args and env variables
+# Replace AWS client zipfile with specific file to pin to a specific version
+# (eg: "awscli-exe-linux-x86_64-2.0.30.zip")
+ARG AWSCLI_VERSION="awscli-exe-linux-x86_64.zip"
+ENV AWSCLI_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}"
```
qhua948, line 9: Shall we install this with EPEL? I think it is better

Dockerfile
```diff
@@ -19,6 +29,18 @@ RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux
...
 RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
 
+# Install aws-cli
+RUN mkdir -p /aws/bin
+WORKDIR /aws
+# Download the awscli zip file
+RUN curl -sSLf $AWSCLI_URL -o awscliv2.zip
+# Extract the awscli zip
+RUN unzip awscliv2.zip
+# Install the libs to the usual location, so the simlinks will be right
```
qhua948, line 39: symlink

scripts/CEE/registry-s3-bucket-size/metadata.yaml
```diff
@@ -0,0 +1,27 @@
+file: script.sh
+name: etcd-health-check
```
sedroche, line 2: You need to update the name here

Dockerfile
karthikperu7: is it available in EPEL? I mixed official docs and ocm-container's method of installation of the aws cli

Dockerfile
karthikperu7: fixed

scripts/CEE/registry-s3-bucket-size/metadata.yaml
karthikperu7: fixed

Dockerfile
```diff
@@ -2,6 +2,16 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS build-stage0
 ARG OC_VERSION="stable"
 ENV OC_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OC_VERSION}"
 
+# AWS cli args and env variables
+# Replace AWS client zipfile with specific file to pin to a specific version
+# (eg: "awscli-exe-linux-x86_64-2.0.30.zip")
+ARG AWSCLI_VERSION="awscli-exe-linux-x86_64.zip"
+ENV AWSCLI_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}"
+ENV AWSSIG_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}.sig"
```
georgettica, line 10: Is the signature used?
I didn't see that in the dockerfile 😳

karthikperu7: No, it is  a copy paste error on my part, good catch. Fixed now :)

scripts/CEE/registry-s3-bucket-size/script.sh
```diff
@@ -0,0 +1,40 @@
...
+
+get_s3_access_secretkey() {
+  local b64enckey=
+  b64enckey=$(oc -n openshift-image-registry get secret image-registry-private-configuration -o jsonpath='{.data.credentials}')
+  echo "${b64enckey}" | base64 --decode |  grep aws_secret_access_key | cut -d= -f2- | tr -d "[:blank:]"
+}
+
+bucket_size() {
+  local bucket="${1}" accesskey="${2}" secretkey="${3}"
+  AWS_ACCESS_KEY_ID="${accesskey}" AWS_SECRET_ACCESS_KEY="${secretkey}" aws s3api list-objects --bucket "${bucket}" --output json --query "sum(Contents[].Size)"
```
georgettica, line 26: is that command region specific? if so, should we pass the  `AWS_DEFAULT_REGION` aswell

Dockerfile
georgettica: I guess this is what you wanted
https://aws.amazon.com/premiumsupport/knowledge-center/ec2-enable-epel/
but I am not sure how good that is

scripts/CEE/registry-s3-bucket-size/script.sh
karthikperu7: I dont think `s3api` command is region specific, eg: https://docs.aws.amazon.com/cli/latest/reference/s3api/list-buckets.html#examples

I also the script on two clusters that were installed in two different regions (`ap-southeast-2` and `us-west-2` )

PR #36, opened 2021-11-24
Add get Kafka instance state script

racheljpg:
This PR is to add a script and accompanying metadata file that can be ran against a single Kafka instance and will then return a tar stream of all the logs/events/CRs in the instance.

### DIFF
--- a/scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
+++ b/scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -o nounset
+set -o pipefail
+
+function main() {
+  local timestamp
+  timestamp=$(date +%s)
+  local inspectDir
+  inspectDir="inspect.${resource_namespace}.${timestamp}"
+  mkdir "${inspectDir}"
+
+  if ! oc get namespace "${resource_namespace}" &> /dev/null; then
+      echo "No namespace found with name ${resource_namespace}" > "${inspectDir}"/failed.txt
+      # Streaming tar.gz so that the output is in the expected format, even in this failure case
+      tar zcf - ./"${inspectDir}"
+      return 1
+  fi
+
+  local managedKafkaCR
+  managedKafkaCR=$(oc -n "${resource_namespace}" get managedkafka -o name)
+  local kafkaCR
+  kafkaCR=$(oc -n "${resource_namespace}" get kafka -o name)
+  local kafkaStatefulSet
+  kafkaStatefulSet=$(oc -n "${resource_namespace}" get statefulset -l app.kubernetes.io/name=kafka -o name)
+
+  oc -n "${resource_namespace}" adm inspect ns/"${resource_namespace}" "${managedKafkaCR}" "${kafkaCR}" \
+    --dest-dir "${inspectDir}" > "${inspectDir}"/ocadm.log 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-topics.sh \
+    --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-topics.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-consumer-groups.sh \
+    --bootstrap-server localhost:9096 --all-groups --describe  > "${inspectDir}"/kafka-consumer-groups.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-acls.sh \
+    --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-acls.txt 2>&1
+
+  find "${inspectDir}" -type f -name secrets.yaml -delete
+
+  find "${inspectDir}" -path '*/managedkafkas/*' -name "*.yaml" -exec sed -i '/password:/d' {} \;
+
+  tar zcf - ./"${inspectDir}"
+}
+
+main
--- a/scripts/managed-services/get-kafka-instance-state/metadata.yaml
+++ b/scripts/managed-services/get-kafka-instance-state/metadata.yaml
@@ -1,0 +1,48 @@
+file: get-kafka-instance-state.sh
+name: get-kafka-instance-state
+description: |
+  Print out the logs for the specified Kafka instance.
+  The following commands will create a directory with all instance information:
+
+  ocm backplane managedjob logs <job-name> > mykafkadebug.tar.gz
+  tar -xvf mykafkadebug.tar.gz
+author: racheljpg
+allowedGroups:
+  - SREP
+  - CSSRE
+  - CEE
+rbac:
+  roles:
+    - namespace: "default"
+      rules:
+        - verbs:
+            - "get"
+            - "list"
+          apiGroups:
+            - ""
+          resources:
+            - "pods/exec"
+        - verbs:
+            - "create"
+          apiGroups:
+            - ""
+          resources:
+            - "pods/exec"
+      clusterRoleRules:
+        - verbs:
+            - "adm"
+          apiGroups:
+            - ""
+          resources:
+            - "inspect"
+        - verbs:
+            - "create"
+          apiGroups:
+            - ""
+          resources:
+            - "pods/exec"
+envs:
+  - key: "resource_namespace"
+    description: "Namespace for the Kafka instance you want to query"
+    optional: false
+language: bash


scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+function main() {
+  local kafkaId=$1
+  local managedKafkaCR=$(oc -n kafka-$kafkaId get managedkafka -o name)
+  local kafkaCR=$(oc -n kafka-$kafkaId get kafka -o name)
+  local kafkaStatefulSet=$(oc -n kafka-$kafkaId get statefulset -l app.kubernetes.io/name=kafka -o name)
+  local timestamp=$(date +%s)
+  local inspectDir="inspect.kafka-$kafkaId.$timestamp"
+  local namespaceCheck=$(oc get namespace kafka-$kafkaId)
```
georgettica, line 10: Why not move it higher? As you are doing checks on Kafka before

scripts/managed-services/get-kafka-instance-state/metadata.yaml
```diff
@@ -0,0 +1,30 @@
+file: get-kafka-instance-state.sh
+name: managed-services/get-kafka-instance-state
+description: Print out the logs for the specified Kafka instance. `oc logs <pod-running-job> | tar -xzf -` will create a directory with all instance information.
```
georgettica, line 3: Won't this fails if any of the commands are not connected as expected?

Also, it's. Good question on how to transfer data between the command outside

I can think of saving a tarfile inside and then adding to the tooling around the script an option to pull a file outside

scripts/managed-services/get-kafka-instance-state/metadata.yaml
```diff
@@ -0,0 +1,30 @@
...
+          resources:
+            - "pods/exec"
+        - verbs:
+            - "create"
+          apiGroups:
+            - ""
+          resources:
+            - "pods/exec"
+envs:
+  - key: "kafkaId"
```
georgettica, line 27: Maybe add something to this script that will show what kafka IDs exist in this cluster (this making the kafkaid optional)

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
...
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9096 --list > ${inspectDir}/kafka-topics.txt 2>&1
+
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server localhost:9096 --all-groups --describe  > ${inspectDir}/kafka-consumer-groups.txt 2>&1
+
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-acls.sh --bootstrap-server localhost:9096 --list > ${inspectDir}/kafka-acls.txt 2>&1
+
+  tar zcf - ./${inspectDir}
+}
+
+main $1
```
georgettica, line 23: This won't hold the kafkaid, the way the arguments work is by adding an envvar with the corresponding name to the pod

georgettica (commented):
Seems nice, I had some changes I suggested and I hope it helps you along with your work

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
k-wall, line 2: enable these bash options (bash best practice)

```
set -e
set -o nounset
set -o pipefail

```

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
...
+function main() {
+  local kafkaId=$1
+  local managedKafkaCR=$(oc -n kafka-$kafkaId get managedkafka -o name)
+  local kafkaCR=$(oc -n kafka-$kafkaId get kafka -o name)
+  local kafkaStatefulSet=$(oc -n kafka-$kafkaId get statefulset -l app.kubernetes.io/name=kafka -o name)
+  local timestamp=$(date +%s)
+  local inspectDir="inspect.kafka-$kafkaId.$timestamp"
+  local namespaceCheck=$(oc get namespace kafka-$kafkaId)
+
+  oc -n kafka-$kafkaId adm inspect ns/kafka-$kafkaId $managedKafkaCR $kafkaCR --dest-dir $inspectDir > /dev/null 2>&1
```
k-wall, line 12: Don't assume that kafka namespace will always be prefixed by 'kafka-'.  Just accept the fully namespace as an argument to the script.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+function main() {
+  local kafkaId=$1
```
SamBarker, line 4: Would it be more user friendly to base things on cluster name?

Maybe use something like 

```bash
local clusterName=$1
local kafkaId=$(oc get pods -A -l strimzi.io/cluster=${clusterName} -o json \
    | jq -r ".items[].metadata.namespace" \
    | head -n1)
```

Which would give:
```bash
❯ echo $clusterName
sbarker-kafka
❯ kafkaId=$(oc get pods -A -l strimzi.io/cluster=${clusterName} -o json | \
 jq -r ".items[].metadata.namespace"  | \
 head -n1)
❯ echo $kafkaId
kafka-c6ens493tq20ii2uemig
```

Which is the full name of the namespace so need for `kafka-$kafkaId` usage.
 
Assuming `jq` is available in the backplane image.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,23 @@
...
+  local namespaceCheck=$(oc get namespace kafka-$kafkaId)
+
+  oc -n kafka-$kafkaId adm inspect ns/kafka-$kafkaId $managedKafkaCR $kafkaCR --dest-dir $inspectDir > /dev/null 2>&1
+
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9096 --list > ${inspectDir}/kafka-topics.txt 2>&1
+
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server localhost:9096 --all-groups --describe  > ${inspectDir}/kafka-consumer-groups.txt 2>&1
+
+  oc exec -n kafka-$kafkaId $kafkaStatefulSet -c kafka -- sh /opt/kafka/bin/kafka-acls.sh --bootstrap-server localhost:9096 --list > ${inspectDir}/kafka-acls.txt 2>&1
+
```
SamBarker, line 19: Possibly out of scope for this script but I think it would be useful to capture the k8s setup when the commands were run.

```
oc get pods -n $kafkaId -o json > ${inspectDir}/namespace.json
```

SamBarker (commented):
Looking good @racheljpg

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
SamBarker: Or is that thought just a duplicate of the `oc adm inspect` ?

scripts/managed-services/get-kafka-instance-state/metadata.yaml
k-wall: CEE will know what you are calling `kafkaId` as `resource_namespace` (from the MK ID Finder).  You should use the same terminology when naming the argument.

scripts/managed-services/get-kafka-instance-state/metadata.yaml
k-wall: The idea was to gather the artefacts (which will comprise kube and kafka artefacts) in an tar archive which willl follow a standard format to ease debugging.  The CEE user of the `ocm  backplane managedjob` command can then direct the job's log to file to get the tgz.

Is there a better way to pull a file from ocm backplane job?

```
ocm  backplane managedjob logs openshift-job-h55l9 > mykafkadebug.tgz
tar xvf mykafkadebug.tgz
```

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: We don't what the output to include the kubermetes secrets.  AFAIK you can't tell `oc adm` not to get secrets, so just prune the secrets directory afterwards:
```
find ${inspectDir} -type d -name secrets -delete
```

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: `ocm adm inspect` captures both logs and the state of the pod resource (as yaml).

```
ind inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0/
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0/
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//mykafka-dev-zookeeper-0.yaml
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper/zookeeper
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper/zookeeper/logs
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper/zookeeper/logs/current.log
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper/zookeeper/logs/previous.log
inspect.local.1420828432188943909/namespaces/kafka-c52ajhf2ek06ri44ife0/pods/mykafka-dev-zookeeper-0//zookeeper/zookeeper/logs/previous.insecure.log```

**Comment by k-wall** [comment]
@georgettica I see the jq is available for use in backplane scripts.  is yq available too?

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 8 above)
Comment (SamBarker) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh:19]: Ignore me then 😁

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 7 above)
Comment (georgettica) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh:4]: Jq is available

**Comment by georgettica** [comment]
@k-wall it's not as of now
https://github.com/openshift/managed-scripts/blob/e1a5bb57afdbcb6f9b89247d63b51fb04d08c377/Dockerfile#L48
You can ask to add it

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 5 above)
Comment (racheljpg) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh:4]: I tested this by scaling down my statefulSet and trying to run the command - it stopped it from running which meant that the logs on the oc exec commands timing out were lost. Removing `set -e` and adding some extra checks has solved that problem. What do you think of the updated version? Should I add `set -e` back, even if it means some logs/info in this situation may be lost?

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 7 above)
Comment (racheljpg) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh]: As far as I'm aware, the CEE engineers will be getting the Kafka instance from something like the `MK - ID Finder dashboard`. @StevenTobin is that correct?

**Comment by racheljpg** [comment]
Hi all, thanks very much for all the feedback and questions. I've incorporated most of it and made some other changes.

One question I have is on the namespace that's set in the metadata file - currently it's set to `redhat-kas-fleetshard-operator`, does that need to be changed? I'm assuming it needs to be set in a namespace that we know will exist, so I'm not sure how I would set it to run in the Kafka instances namespace?
Thanks!

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 7 above)
Comment (StevenTobin) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh]: Kafka cluster names are not unique so the id is the reliable way to identify which kafka cluster we want to target.

File: scripts/managed-services/get-kafka-instance-state/metadata.yaml [hunk 9]
```diff
@@ -0,0 +1,42 @@
...
+  ocm backplane managedjob logs <job-name> > mykafkadebug.tgz
+  tar -xvf mykafkadebug.tgz
+author: racheljpg
+allowedGroups:
+  - SREP
+  - CSSRE
+  - CEE
+rbac:
+  roles:
+    - namespace: "redhat-kas-fleetshard-operator"
```
k-wall, line 16: This doesn't look right.  This script is nothing to do with the fleetshard operator, so it doesn't seem right to use that operator's namespace.  Also there a practical problem, the name of this namespace differs between stage and production redhat-kas-fleetshard-operator and redhat-kas-fleetshard-operator-qe.

Can we use the default namespace?

@StevenTobin could I get your opinion on this please?

scripts/managed-services/get-kafka-instance-state/metadata.yaml
```diff
@@ -0,0 +1,42 @@
...
+          resources:
+            - "pods/exec"
+        - verbs:
+            - "create"
+          apiGroups:
+            - ""
+          resources:
+            - "pods/exec"
+      clusterRoleRules:
+        - verbs:
```
k-wall, line 32: The script needs to `oc exec <pod>` from any kafka instance namespace.  We don't know the names of the namespaces as they will be defined in advance.  I would have thought we'd have need cluster role permission.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: On reflection, this script is okay without the `-e` flag.

scripts/managed-services/get-kafka-instance-state/metadata.yaml
StevenTobin: Yeah @racheljpg and I discussed this, ideally it would run in the namespace of the kafka cluster you are targeting but failing that I recommended a namespace that we create but as you rightly point out the namespace difference between prod and stage is a problem for that.

I don't think we can specify the kafka namespace since that is specified as a variable but I would love confirmation on that.

I would be fine with it running in the default namespace.

k-wall: > I don't think we can specify the kafka namespace since that is specified as a variable but I would love confirmation on that.

For that to work, the rbac rules would have to support parameterization.  I wouldn't put money on ocm backplane supporting that.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,45 @@
...
+  local kafkaStatefulSet
+  kafkaStatefulSet=$(oc -n "${resource_namespace}" get statefulset -l app.kubernetes.io/name=kafka -o name)
+  local timestamp
+  timestamp=$(date +%s)
+  local inspectDir
+  inspectDir="inspect.${resource_namespace}.${timestamp}"
+
+  mkdir "${inspectDir}"
+
+  if [[ ! $(oc get namespace "${resource_namespace}") ]]; then
```
k-wall, line 19: Move this defensive code up to before you first use ${resource_namespace}, i.e. line6.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,45 @@
...
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-consumer-groups.sh \
+    --bootstrap-server localhost:9096 --all-groups --describe  > "${inspectDir}"/kafka-consumer-groups.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-acls.sh \
+    --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-acls.txt 2>&1
+
+  find "${inspectDir}" -type f -name secrets.yaml -delete
+
+  sed -i  '/password:/d' "$(find "${inspectDir}" -path '*/managedkafkas/*'  -name "*.yaml")"
```
k-wall, line 40: "SC2046: Quote this to prevent word splitting" advice for this line is actually incorrect in this case.
See https://github-wiki-see.page/m/koalaman/shellcheck/wiki/SC2046 for the alternative formulation that will avoid the nag from shellcheck.

racheljpg:
@georgettica I've made some changes to the script if you wouldn't mind reviewing. Thanks!

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,46 @@
...
+set -o pipefail
+
+function main() {
+  local timestamp
+  timestamp=$(date +%s)
+  local inspectDir
+  inspectDir="inspect.${resource_namespace}.${timestamp}"
+  mkdir "${inspectDir}"
+
+  if [[ ! $(oc get namespace "${resource_namespace}" &> /dev/null) ]]; then
```
georgettica, line 12: This is the only place I think it'll contaminate the tarfile

racheljpg: Should that still happen, even with the new `&> /dev/null` addition?

qhua948: I think what you are looking for is the exit code right?

Then it sohuld be `if oc get namespace "${resource_namespace}" &> /dev/null; then`

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -0,0 +1,46 @@
...
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-consumer-groups.sh \
+    --bootstrap-server localhost:9096 --all-groups --describe  > "${inspectDir}"/kafka-consumer-groups.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-acls.sh \
+    --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-acls.txt 2>&1
+
+  find "${inspectDir}" -type f -name secrets.yaml -delete
+
+  read -ra args < <(find "${inspectDir}" -path '*/managedkafkas/*' -name "*.yaml")
```
qhua948, line 40: Can you use `find "${inspectDir}" -path '*/managedkafkas/*' -name "*.yaml" -exec sed  -i '/password:/d' {} \;`instead here?

k-wall:
@racheljpg if you can add me https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-access-to-your-personal-repositories/inviting-collaborators-to-a-personal-repository i can address the feedback

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
georgettica: I didn't see the `&>` when I reviewed.. I'll look again :)

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: Done.

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: Done

scripts/managed-services/get-kafka-instance-state/metadata.yaml
```diff
@@ -0,0 +1,48 @@
+file: get-kafka-instance-state.sh
+name: managed-services/get-kafka-instance-state
+description: |
+  Print out the logs for the specified Kafka instance.
+  The following commands will create a directory with all instance information:
+
+  ocm backplane managedjob logs <job-name> > mykafkadebug.tgz
```
georgettica, line 7: As all of the script looks correct, making sure that .tgz to the correct file extension

Isn't it .tar.gz?
It would be good to test this before merging

scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh
StevenTobin: I don't think this check is working correctly

```
$ oc get namespace kafka-c6f606rg996p1r6q4kj0
NAME                         STATUS   AGE
kafka-c6f606rg996p1r6q4kj0   Active   46h
$ oc get namespace kafka-c6f606rg996p1r6q4kj0 &> /dev/null
$ echo $?
0
```
I also tried running this script using the new testing tooling and it's failing on this check even when I confirm it can see the namespace.

`if ! oc get namespace "${resource_namespace}" &> /dev/null; then` is working correctly because we want to fail out if it cannot find the namespace.

scripts/managed-services/get-kafka-instance-state/metadata.yaml
```diff
@@ -0,0 +1,48 @@
+file: get-kafka-instance-state.sh
+name: managed-services/get-kafka-instance-state
```
StevenTobin, line 2: This is causing a problem when running with the testing tooling

```
ERRO[0004] error from backplane: 
 Status Code: 500
 Message: Unable to create test job on the cluster, error: invalid canonical name, can't have more than 2 slashes
 ```
 I think it should just be `get-kafka-instance-state`

Comment (racheljpg) [reply scripts/managed-services/get-kafka-instance-state/metadata.yaml]: I've changed that now

File: scripts/managed-services/get-kafka-instance-state/metadata.yaml (hunk 15 above)
Comment (racheljpg) [reply scripts/managed-services/get-kafka-instance-state/metadata.yaml]: Thanks, I've changed that now

**Comment by georgettica** [comment]
Basically
/lgtm
But I would also 
/hold
As the error that was provided was not solved
Regarding https://github.com/openshift/managed-scripts/pull/36#discussion_r757535903

**Comment by k-wall** [comment]
@georgettica the error @StevenTobin reported should be been resolved.

**Comment by georgettica** [comment]
if so, let's give it a go!
but before can we squash commits?
I just saw 3 `addressing review feedback` ones, so maybe there is more squashing to do

**Comment by racheljpg** [comment]
@georgettica I've squashed the other commits now! Thanks :)

**Comment by georgettica** [comment]
/approve
🤞

**Comment by racheljpg** [comment]
@georgettica it does not seem to have picked up the "lgtm" label

File: scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh (hunk 13 above)
Comment (qhua948) [reply scripts/managed-services/get-kafka-instance-state/get-kafka-instance-state.sh]:
```
 edward@derg-arch  ~  oc get ns asdf
echo $?
Error from server (NotFound): namespaces "asdf" not found
 ✘ edward@derg-arch  ~  echo $?
1
 edward@derg-arch  ~ 
```

PR #39, opened 2021-11-29
Fix #37: Include yq in the managed script image

### LINKED ISSUE #37: Add yq to command line tools
In our use-case (Managed Kafka) we are using `oc adm inspect` to grab kubernetes state.  In some cases, we want to exclude some fields from the yaml output files generated by `inspect`.    Unfortunately the `oc adm inspect` command line does not support a `--output=json` mode, so we cannot use `jq`.

Could you consider adding `yq` to the command line tools available to an ocm backplace script?

Happy to raise a PR.

k-wall:


georgettica:
I saw the addition, but it seems the `yq` package is not found 😱

k-wall:
Closing until such time `yq` is available in the UBI yam repo.

PR #41, opened 2021-12-02
Add RBAC for managedkafka and kafka resources

k-wall:
The script requires these RBAC permissions to lookup the resource name (`oc get ...`).

PR #42, opened 2021-12-06
Add missing permissions to kafka-instances script

k-wall:
statefulset and pod permissions were absent, this caused the commands that query kafka state to fail.

### DIFF
--- a/scripts/CEE/get-kafka-instance-state/metadata.yaml
+++ b/scripts/CEE/get-kafka-instance-state/metadata.yaml
@@ -42,6 +42,19 @@ rbac:
           resources:
             - "pods/exec"
         - verbs:
+            - "get"
+            - "list"
+          apiGroups:
+            - "apps"
+          resources:
+            - "statefulsets"
+        - verbs:
+            - "get"
+          apiGroups:
+            - ""
+          resources:
+            - ""
+        - verbs:
             - "list"
           apiGroups:
             - "managedkafka.bf2.org"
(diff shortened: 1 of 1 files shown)


georgettica:
nod to last time the checks were made
https://github.com/openshift/managed-scripts/pull/41

what was missing there that you caught now?

k-wall:
yes, indeed, sorry for the churn on this.

For this mistake to have slipped through there must have been some mistake made with the application of the testing procedure when rachel did the original work.    She was using https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts    "How to test the script before creating the PR." .  I'm unsure why she didn't see these errors. 
 
In addition we are still having problems running `ocm backplane testjobs` which we are still working through separately. (@StevenTobin )

scripts/CEE/get-kafka-instance-state/metadata.yaml
```diff
@@ -41,6 +41,19 @@ rbac:
...
           resources:
             - "pods/exec"
+        - verbs:
+            - "get"
+            - "list"
+          apiGroups:
+            - "apps"
+          resources:
+            - "statefulsets"
+        - verbs:
```
georgettica, line 51: Isn't that redundant / too permissive?
Why do you need this section?

georgettica:
Hopefully the last 

But I think I found a permission issue I notes
After that I'll lgtm

scripts/CEE/get-kafka-instance-state/metadata.yaml
k-wall: thanks, well spotted.  I made a mistake copying back from the rbac yaml test case into the metadata.yml.

PR #45, opened 2021-12-14
Migrating the list-alerts script into managed-scripts

bdematte:
Migrating the list-alerts script into managed-scripts. Doing so, taking also the opportunity to allow CEE to proceed with the 'read-only' actions such as listing alerts and/or silences.


georgettica:
Can we maybe put this in a general SREP+CEE folder?

scripts/CEE/list-alerts/script.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "--pending-only, --firing-only and --all are mutually exclusive."
+    exit 2
+fi
+
+if [[ $((IS_WARNING_ONLY+IS_CRITICAL_ONLY)) -gt 1 ]] ; then
+    echo "--warning-only and --critical-only are mutually exclusive."
+    exit 2
+fi
+
+if [[ $IS_ALL_STATES == 1 ]] ; then
```
georgettica, line 55: Could you use the firing-only flag instead?
I like that format more than:
IF all
ELIF one
ELSE two
---
Format I like
IF one
ELIF two
Else all

scripts/CEE/list-alerts/metadata.yaml
```diff
@@ -0,0 +1,34 @@
+file: script.sh
+name: list-alerts
+description: Script listing the active/pending alerts in alertmanager. It also provide an option to list active silences. 
+author: bdematte
+allowedGroups: 
+  - SREP
```
georgettica, line 6: If this has all the groups, why do we need to duplicate it?

bdematte:
> Can we maybe put this in a general SREP+CEE folder?

The whole CEE folder is SREP+CEE (any script that can be ran by CEE can as well be ran by SREP and we don't want to dupplicate it). So this is more a matter for the repo in general but not specific to this MR I believe.
Imho, in term of usage, you would list the available scripts with `ocm backplane scripts list` to get the scipts you can personally use and not scroll/browse in this repository, I don't think this would make a difference.

scripts/CEE/list-alerts/script.sh
bdematte: I think it is more logical to have the defaults behavior in the 'else' statement. 
IS_ALL_STATES and IS_PENDING_ONLY are markers for options being used, and the else statement covers 'all other cases' which can be default or usage of the explicit option.

scripts/CEE/list-alerts/metadata.yaml
bdematte: This is probably something for the repos owner and not something specific to this PR but more on how managed-scripts are working. 
Probably this is a discussion to have in #sd-ism-backplane and card the implementation if we want to change it (even though, I am not sure all scripts will have necessarily the same groups in a given folder).

georgettica:
that's also ok.
if we remove the script from the SREP folder it's ok by me aswell

bdematte:
> that's also ok. if we remove the script from the SREP folder it's ok by me aswell
> 
> as long as there are no duplicates i'm happy

The script added to SREP doesn't do the same actions : 
 * CEE one is managing the read only actions (list alerts and silence)
 * SREP one is managing the write actions (create/delete silence). 
The only dupplication is the list silence feature because it is useful to be able to trigger it right after delete/create ran to check for it.

scripts/CEE/list-alerts/script.sh
bng0y: I'm good with this as long as the behavior is documented (which is the case)

scripts/SREP/manage-silence/script.sh
```diff
@@ -0,0 +1,185 @@
...
+AM_HOST=$(_get_host alertmanager-main)
+
+if [[ $CREATE_SILENCE == 1 ]] ; then
+    if [[ -z $USERNAME ]] ; then
+        echo "Aborting. Username needs to be provided when setting a silence"
+        exit 1
+    fi 
+
+    prompt_for_silence_info
+    curl -s -k -X POST -H "Content-Type: application/json" --data "$SILENCE_JSON" -H "Authorization: Bearer $PROM_TOKEN" "https://$AM_HOST/api/v1/silences"
```
bng0y, line 157: can we use the response to validate that the silence was indeed created?

scripts/CEE/list-alerts/README.md
```diff
@@ -0,0 +1,39 @@
+# list-alerts script
+
+## Purpose
+
+The script is used to ease liar on-going alerts and silence for a given cluster.
```
fahlmant, line 5: I'm not sure what this sentence is saying

fahlmant (approved):
/lgtm

PR #46, opened 2021-12-15
Add optional since/since-time flags as environment variables

racheljpg:
This is a PR to make an edit to the get-kafka-instance-state script to allow optional since/since-time flags to be specified.

### DIFF
--- a/scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
+++ b/scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -o pipefail
+set -o nounset
 
 function main() {
   local timestamp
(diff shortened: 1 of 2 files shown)


scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -o nounset
```
k-wall, line 2: using -nounset is best practice, so we don't favour turning it off.
use the idiom described [here](https://stackoverflow.com/questions/7832080/test-if-a-variable-is-set-in-bash-when-using-set-o-nounset/9824943) instead.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -23,8 +22,16 @@ function main() {
   local kafkaStatefulSet
   kafkaStatefulSet=$(oc -n "${resource_namespace}" get statefulset -l app.kubernetes.io/name=kafka -o name)
 
-  oc -n "${resource_namespace}" adm inspect ns/"${resource_namespace}" "${managedKafkaCR}" "${kafkaCR}" \
-    --dest-dir "${inspectDir}" > "${inspectDir}"/ocadm.log 2>&1
+  if [[ -v since ]]; then
```
k-wall, line 25: https://deviq.com/principles/dont-repeat-yourself applies.
Having to repeat the same thing over and over, changing one thing is a bad smell.   Always look for a better way to solve the problem.

bash supports arrays... and appending to arrays

arrVar=("AC" "TV" "Mobile" "Fridge" "Oven" "Blender")
arrVar+=("Dish Washer")
echo "${arrVar[@]}"

so how about creating an array first with all your common arguments, then use conditional logic append either since or since_time arguments parts.  Finally execute passing the array

```
oc -n "${resource_namespace}" adm inspect "${arrVar[@]}"
```

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
racheljpg: I've added `set -o nounset` back, but I think because of the removal/change to the if statement, there are no longer any unset variables anyway - am I misunderstanding that?

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
racheljpg: Thank you Keith, that's much more efficient. I've pushed some changes now!

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: correct, change looks good. okay to resolve comment.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
k-wall: change looks good.  pushed one more change to include --dest-dir in inspectOps too. okay to resolve comment.

georgettica:
Sorry I didn't see your message
/retest

georgettica:
/lgtm
/approve
Looks good and checks pass

PR #49, opened 2022-01-21
renamed matadata.yml to metadata.yaml

supreeth7:


georgettica:
@wanghaoran1988 we should card and address this maybe in a wider level?

supreeth7:
> we should card and address this maybe in a wider level?

I agree, a managedJob should be created regardless of the metadata file extension

PR #50, opened 2022-01-21
OSD-8688 Add cluster-operator-status script

mrbarge:
Refs: [OSD-8688](https://issues.redhat.com/browse/OSD-8688)

This PR adds a `cluster-operator-status` managed script script to gather data on degraded or unavailable cluster operators.

It can optionally target a specific cluster operator via the `CLUSTER_OPERATOR` parameter.

Successfully tested via `ocm backplane testjob create`. (An easy scenario to test against that quickly simulates cluster operator failures is to cordon and drain a master node.)

### DIFF
--- a/scripts/SREP/cluster-operator-status/script.sh
+++ b/scripts/SREP/cluster-operator-status/script.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+# This script provides a rich summary of information concerning the status of 
+# degraded/unavailable OpenShift cluster operators.
+#
+# The script can specifically target a cluster operator through
+# the use of a "CLUSTER_OPERATOR" parameter
+#
+# The script can adjust the time interval of logging retrieved through the
+# use of the OC_LOGS_SINCE parameter. This parameter can define
+# additional options supplied to 'oc logs' when gathering pod logs.
+#
+# The script can also filter what logging is retrieved through the
+# use of the LOG_PATTERN parameter. This parameter defines an 
+# extended regular expression passed to grep to filter logs
+# after being returned from 'oc logs'
+#
+set -e
+set -o nounset
+set -o pipefail
+
+
+# print_header prints a formatted section break header
+print_header() {
+    clusteroperator=$1
+    heading=$2
+    echo -e "--- [ CLUSTER OPERATOR '${clusteroperator}' ${heading} ] ---\n"
+}
+
+# get_summary prints the basic status for the given cluster operator parameter
+get_summary() {
+    clusteroperator=$1
+
+    # shellcheck disable=SC2016
+    available=$(jq -r --arg CLUSTER_OPERATOR "${clusteroperator}" '.items[] | select(.metadata.name == $CLUSTER_OPERATOR) | .status.conditions[] | select(.type == "Available") | "\(.status) --> \(.message)"' <<< "$OPERATOR_REPORT")
+    # shellcheck disable=SC2016
+    progressing=$(jq -r --arg CLUSTER_OPERATOR "${clusteroperator}" '.items[] | select(.metadata.name == $CLUSTER_OPERATOR) | .status.conditions[] | select(.type == "Progressing") | "\(.status) --> \(.message)"' <<< "$OPERATOR_REPORT")
+    # shellcheck disable=SC2016
+    degraded=$(jq -r --arg CLUSTER_OPERATOR "${clusteroperator}" '.items[] | select(.metadata.name == $CLUSTER_OPERATOR) | .status.conditions[] | select(.type == "Degraded") | "\(.status) --> \(.message)"' <<< "$OPERATOR_REPORT")
+
+    print_header "${clusteroperator}" "SUMMARY"
+    echo " - AVAILABLE:   ${available}"
+    echo " - DEGRADED:    ${degraded}"
+    echo " - PROGRESSING: ${progressing}"
+    echo -e "\n"
+}
+
+# get_description prints a detailed status description for the given cluster 
+# operator parameter
+get_description() {
+    clusteroperator=$1
+
+    if ! co_desc_report=$(oc describe clusteroperator "${clusteroperator}"); then
+       echo "An error was returned when describing ClusterOperator '${clusteroperator}': ${co_desc_report}"
+       return
+    fi
+    print_header "${clusteroperator}" "DESCRIPTION"
+    echo -e "${co_desc_report}\n\n"
+}
+
+# get_logs prints the last 15 lines of logs from the cluster operator pod
+get_logs() {
+    clusteroperator=$1
+    # Look up all namespaces referenced in status
+    # shellcheck disable=SC2016
+    relatedns=$(jq -r --arg CLUSTER_OPERATOR "${clusteroperator}" '.items[] | select(.metadata.name == $CLUSTER_OPERATOR) | .status.relatedObjects[] | select(.resource == "namespaces") | .name' <<< "$OPERATOR_REPORT")
+    while IFS= read -r ns; do
+        ns_report=$(oc get deployment --ignore-not-found -n "$ns" -o name)
+        while IFS= read -r deployment; do
+            # Look for any deployment which matches an operator deployment.
+            # The cluster operator name may or may not have the word operator in it.
+            # The cluster operator namespaces may or may not unrelated operator pods in them.
+            if [[ "${deployment}" =~ ${clusteroperator} && "${deployment}${clusteroperator}" =~ operator ]]; then
+                # deploymentlogs=$(log_cmd "${deployment}" "${ns}")
+                if [[ -n "${LOG_PATTERN}" ]]; then
+                    # shellcheck disable=SC2046
+                    deploymentlogs=$(grep -iE "${LOG_PATTERN}" <<< $(oc logs "${deployment}" -n "${ns}" --all-containers "${OC_LOGS_SINCE}") || true)
+                else 
+                    deploymentlogs=$(oc logs "${deployment}" -n "${ns}" --all-containers "${OC_LOGS_SINCE}")
+                fi
+                print_header "${clusteroperator}" "DEPLOYMENT LOGS: ${deployment}"
+                echo -e "${deploymentlogs}\n\n"
+            fi
+        done <<< "${ns_report}"
+    done <<< "${relatedns}"
+}
+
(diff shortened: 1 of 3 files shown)


georgettica:
I was looking at the PR but it seems there are build issues..
Ping me when your build is passing

mrbarge:
@georgettica after much wrangling with false positives from the old ShellCheck (that don't fire in 0.7.3) I've gotten the build to work. Thanks for your time in having a look.

@feichashao as the creator of the original ticket if you could also see if it aligns with what you were hoping to see that would be much appreciated, thank you. I elected to go with 'last 15 lines' of logs rather than 'last 5 minutes' as for some operators I found that 'last 5 minutes' can be a LOT of data.

georgettica:
@mrbarge @feichashao we might want to upgrade shellcheck if this is the case

I will TAL

georgettica:
/lgtm
/approve
/hold
As the original ticket creator needs to approve the `oc logs`

georgettica:
Good job regardless 🤯🌟✨💫😽🎉😻

scripts/SREP/cluster-operator-status/script.sh
```diff
@@ -0,0 +1,114 @@
...
+    echo " - PROGRESSING: ${progressing}"
+    echo -e "\n"
+}
+
+# get_description prints a detailed status description for the given cluster 
+# operator parameter
+get_description() {
+    clusteroperator=$1
+
+    if ! co_desc_report=$(oc describe clusteroperator "${clusteroperator}"); then
```
ravitri, line 43:
```suggestion
    if ! co_desc_report=$(oc get co "${clusteroperator}" -ojson | jq -r '.status.conditions[] | select ((.type == "Degraded" or .type == "Unavailable") and (.status == "True")) | [.type, .message]'); then
```

Though the `oc describe` of the resource would have more details, I wonder if we can directly highlight the event message which the CO would update in the `.status.conditions` list. Else, another option can be to keep the `oc describe` in place as it is and have another section with "hints" of what the error could be? WDYT?

scripts/SREP/cluster-operator-status/script.sh
```diff
@@ -0,0 +1,114 @@
...
+       echo "An error was returned when describing ClusterOperator '${clusteroperator}': ${co_desc_report}"
+       return
+    fi
+    print_header "${clusteroperator}" "DESCRIPTION"
+    echo -e "${co_desc_report}\n\n"
+}
+
+# get_logs prints the last 15 lines of logs from the cluster operator pod
+get_logs() {
```
ravitri, line 52: Right now, the logs are mostly collected for the operator pod. In case of Monitoring stack and the machine-api CO, several pods are present and each can have multiple containers in it. What are your thoughts on showing logs for all containers too? Else, another option can be to have it for the detailed SOP for the respective CO as well.

scripts/SREP/cluster-operator-status/script.sh
georgettica: I would add the `oc get co` after the report is printed (line 48)

scripts/SREP/cluster-operator-status/script.sh
georgettica: the standard of the container name that usually does the operations is that it's name is the same name as the deployment

this code can also be tested on the monitoring CO and see if it fails with the current implementation

georgettica (commented):
good reviews @ravitri! commented on them a bit to try and explain / direct @mrbarge on solving them

scripts/SREP/cluster-operator-status/script.sh
```diff
@@ -0,0 +1,114 @@
...
+    # shellcheck disable=SC2016
+    relatedns=$(jq -r --arg CLUSTER_OPERATOR "${clusteroperator}" '.items[] | select(.metadata.name == $CLUSTER_OPERATOR) | .status.relatedObjects[] | select(.resource == "namespaces") | .name' <<< "$OPERATOR_REPORT")
+    while IFS= read -r ns; do
+        ns_report=$(oc get deployment --ignore-not-found -n "$ns" -o name)
+        while IFS= read -r deployment; do
+            # Look for any deployment which matches an operator deployment.
+            # The cluster operator name may or may not have the word operator in it.
+            # The cluster operator namespaces may or may not unrelated operator pods in them.
+            if [[ "${deployment}" =~ ${clusteroperator} && "${deployment}${clusteroperator}" =~ operator ]]; then
+                deploymentlogs=$(oc logs "${deployment}" -n "${ns}" --tail=15)
```
ravitri, line 65: 1. For now, the default is last 15 lines. It may or may not be enough to convey the issue. Reason being, the CO degraded alert may fire after 10-20 minutes and the logs might differ for operators. What about adding an environment variable such as `LOG_FLAGS` which the end user can customise and grep as required if the last 15 lines aren't helpful? Example:
```bash
deploymentlogs=$(oc logs "${deployment}" -n "${ns}" "${LOG_FLAGS}")
```

2. Something which can be extended further is to parse for certain common strings for any issue. Example, strings like "error", "fatal", "timeout", "fail" etc. convey a possible hint from the logs. In this way, we could have `--since=5m` logs and `grep -iE "${LOG_PATTERN}"` where `LOG_PATTERN` could be some strings the user expects. Though, this is heavily dependent on what a particular CO logs but has chances of increasing relevance in output by script. Thoughts?

georgettica: but that the flag will have a default of `--tail=15`

scripts/SREP/cluster-operator-status/script.sh
mrbarge: This makes sense, what I ended up doing was add it to the Summary section at the start to print out the `message` associated with each condition. It now behaves like this:

```
--- [ CLUSTER OPERATOR 'authentication' SUMMARY ] ---                                                                                                                                                                               
                                                                                                                                                                                                                                    
 - AVAILABLE:   True --> All is well
 - DEGRADED:    True --> APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable for apiserver.openshift-oauth-apiserver ()
OAuthServerDeploymentDegraded: 1 of 3 requested instances are unavailable for oauth-openshift.openshift-authentication ()
 - PROGRESSING: False --> AuthenticatorCertKeyProgressing: All is well
```

scripts/SREP/cluster-operator-status/script.sh
mrbarge: I've added `--all-containers` to the `oc logs` call now.

scripts/SREP/cluster-operator-status/script.sh
mrbarge: Yeah, this is a good suggestion, thanks. I've implemented this now. 
`LOG_PATTERN` will be appended onto the `oc logs` call. This lets you do things such as

`LOG_PATTERN="--since=5m"`
or throw a pipe on the end, ie.
`LOG_PATTERN="| grep -i error"`

If not specified at all, it will default to `--tail 15`.

mrbarge: On reflection, I'm a little bit nervous about my current implementation which essentially allows anyone to potentially supply an arbitrary executable command via `LOG_PATTERN`, that's ripe for potential abuse. I'll rework that tomorrow into two separate parameters, an `OC_LOGS_ARGS` and a `GREP_PATTERN` which is explicitly passed to `grep`.

scripts/SREP/cluster-operator-status/script.sh
ravitri: Thanks a lot! Yeah, this works too. Basically now the summary would give a glimpse of the issue at the beginning itself. If the user wants to know details then can refer to the logs etc.

scripts/SREP/cluster-operator-status/script.sh
mrbarge: Made another attempt at this.

- `OC_LOGS_SINCE` can be used to define the `since` interval (eg. `30m`). If not specified, last 15 lines of logs are returned. I experimented with allowing generic args (eg. `--since=30m`) be supplied as parameters but backplane seemed to have a problem with argument values that are, in turn, `name=value` pairs - broke its parsing (something I'll follow up)
- `LOG_PATTERN` defines a regular expression that will be used in a grep of the logs returned. If not specified, no grepping is performed. The grepping is done in a safe manner that shouldn't allow sneaky "I'll just slide this command into your pipeline" sort of tricks.

scripts/SREP/cluster-operator-status/script.sh
```diff
@@ -0,0 +1,137 @@
...
+        ns_report=$(oc get deployment --ignore-not-found -n "$ns" -o name)
+        while IFS= read -r deployment; do
+            # Look for any deployment which matches an operator deployment.
+            # The cluster operator name may or may not have the word operator in it.
+            # The cluster operator namespaces may or may not unrelated operator pods in them.
+            if [[ "${deployment}" =~ ${clusteroperator} && "${deployment}${clusteroperator}" =~ operator ]]; then
+                # deploymentlogs=$(log_cmd "${deployment}" "${ns}")
+                if [[ -n "${LOG_PATTERN}" ]]; then
+                    # shellcheck disable=SC2046
+                    deploymentlogs=$(grep -E "${LOG_PATTERN}" <<< $(oc logs "${deployment}" -n "${ns}" --all-containers "${OC_LOGS_SINCE}") || true)
```
ravitri, line 76:
```suggestion
                    deploymentlogs=$(grep -iE "${LOG_PATTERN}" <<< $(oc logs "${deployment}" -n "${ns}" --all-containers "${OC_LOGS_SINCE}") || true)
```
Using `-E` might be more strict in having exact patterns to be matched where the user is expected to know the error strings. Would it make sense to accept strings by ignoring case? Else another option is to have `GREP_FLAGS` env variable but don't want to add too many flags and knobs for users in the start. Am ok with `-E` too but then we could document the defaults so that the user is aware that the strings are case sensitive.

mrbarge: I've changed it to `-iE`

feichashao:
> I elected to go with 'last 15 lines' of logs rather than 'last 5 minutes' as for some operators I found that 'last 5 minutes' can be a LOT of data.

The 'last 5 minutes' is just an example. Printing the last 15 lines is more desirable as it avoids flooding the output. Thank you for also providing grep & since options!

PR #54, opened 2022-02-07
WIP RHOSAK managed script

R-Lawton:


georgettica:
@clcollins @R-Lawton as this PR might have exposed docker logs (means we need to fail to commit if files like this exist?

R-Lawton:
> @clcollins @R-Lawton as this PR might have exposed docker logs (means we need to fail to commit if files like this exist?
Apologies Removed the output and rebased

georgettica (commented):
Needs some more polish and maybe to make sure the code compiles and you will be on track for merging

R-Lawton:
Thank you for the early review i will make those changes but its still a WIP and not ready yet

georgettica:
@R-Lawton to make it clearer, you can set the PR as a draft
You're welcome

PR #65, opened 2022-04-04
updating rbac cluster roles

R-Lawton:
fixing the rbac permissions in the get kafka instance script to remove rbac as the default cluster read permissions is enough  https://issues.redhat.com/browse/MGDSTRM-8074


georgettica:
Great!
/approve
/lgtm
/hold
In case it's ready for review but not to.merge

PR #70, opened 2022-05-23
pcap collector job

sedroche:
https://issues.redhat.com/browse/OSD-11526

Replacing https://github.com/openshift/ops-sop/blob/master/v4/utils/pcap-collector.sh which doesn't work with backplane into a managedjob

### DIFF
--- a/scripts/CEE/pcap-collector/script.sh
+++ b/scripts/CEE/pcap-collector/script.sh
@@ -12,7 +12,20 @@ fi
 #VARS
 NS="openshift-backplane-managed-scripts"
 OUTPUTFILE="/tmp/capture-${NODE}.pcap"
-PODNAME="pcap-collector"
+PODNAME="pcap-collector-${NODE}"
+
+NODES=$(oc get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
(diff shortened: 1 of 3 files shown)


scripts/CEE/pcap-collector/script.sh
```diff
@@ -0,0 +1,55 @@
...
+EOF
+
+while [ "$(oc get pod -n ${NS} ${PODNAME} -o jsonpath='{.status.phase}' 2>/dev/null)" != "Succeeded" ];
+do
+  sleep 1
+done
+
+oc -n $NS logs $PODNAME | gunzip > "$OUTPUTFILE"
+oc -n $NS delete pod $PODNAME >/dev/null 2>&1
+gzip "$OUTPUTFILE" --stdout
```
feichashao, line 55: what's the reason behind doing a gunzip+gzip ?

sedroche: I didn't know how to pipe it to stdout otherwise. I tried to cat it but it didn't work. I'd like to not gunzip+gzip but I dont have the bash skills. Any siuggestions?

fahlmant (commented):
@sedroche This generally looks good to me. Whats the timeline on getting this merged and pushed out for MCS to use?

PR #74, opened 2022-06-16
Removing georgettica from OWNERS -- OHSS-13013

Nikokolas3270:


Nikokolas3270:
@supreeth7 : could you also approve so the PR get merged? Thanks!

PR #75, opened 2022-06-22
Updating the sleep for less api calls

sedroche:


sedroche:
Thanks @wanghaoran1988

PR #76, opened 2022-06-29
OSD-12077 - Add short description to few scripts and update descripti…

Tafhim:
Resolves [OSD-12077](https://issues.redhat.com//browse/OSD-12077)

Added Short Descriptions where long descriptions were not clear when trimmed
Updated some long descriptions to keep them consistent with the rest of the script descriptions and shortened where possible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.swp
 __pycache__/
 *.pyc
+/scripts/**/metadata.json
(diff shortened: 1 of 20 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 *.swp
 __pycache__/
 *.pyc
+/scripts/**/metadata.json
```
wanghaoran1988, line 4: Why we need add this?

wanghaoran1988: ah, I miss read it, that's json files not the yaml ones.

PR #78, opened 2022-08-05
New backplane script to do rolling restarts on Kafka statefulsets or …

pat-cremin:
…individual pods

### DIFF
--- a/scripts/CSSRE/rolling-restart-brokers/README.md
+++ b/scripts/CSSRE/rolling-restart-brokers/README.md
@@ -1,18 +1,18 @@
 # get-kafka-topics
 
-This scripts runs the `rolling-restart-brokers.sh` script to manually roll the Kafka brokers.
+This scripts runs the `rolling-restart-brokers.sh` script to manually roll the Kafka brokers. Up to Strimzi v0.32.x both Statefulsets and Strimzipodsets may be enabled. Beyond v0.32.x only Strimzipodsets will work. 
 
 ## Usage
 
 ### Running locally
 
 ```bash
-Usage: ./rolling-restart-brokers.sh -n KAFKA_NAMESPACE [-s STATEFUL_SET | -p POD]
+Usage: ./rolling-restart-brokers.sh -n KAFKA_NAMESPACE [-s STATEFULSET|STRIMZIPODSET | -p POD]
 
 # Examples
-./rolling-restart-brokers.sh -n kafka-ID
-./rolling-restart-brokers.sh -n kafka-ID -s zookeeper
-./rolling-restart-brokers.sh -n kafka-ID -p broker-kafka-0
+./rolling-restart-brokers.sh -n kafka-namespace
+./rolling-restart-brokers.sh -n kafka-namespace -s zookeeper
+./rolling-restart-brokers.sh -n kafka-namespace -p broker-kafka-0
 ```
 
 ### Running as a managed script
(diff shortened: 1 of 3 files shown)


scripts/CSSRE/rolling-restart-brokers/metadata.yaml
```diff
@@ -0,0 +1,43 @@
...
+        - ""
+      resources:
+        - "pods"
+    - verbs:
+        - "list"
+      apiGroups:
+        - "kafka.strimzi.io"
+      resources:
+        - "kafkas"
+    -verbs:
```
vinaybommana, line 30: Hey @pat-cremin, I think this is where the pipeline is failing, please add space between `- verbs`

scripts/CSSRE/rolling-restart-brokers/metadata.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - CSSRE
+rbac:
+  clusterRoleRules:
+    - verbs:
+        - "get"
+        - "list"
+      apiGroups:
+        - "apps"
+      resources:
+        - "statefulsets"
```
wanghaoran1988, line 15: Why it's clusterRoleRules here? can it be namespace scoped?

scripts/CSSRE/rolling-restart-brokers/metadata.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - verbs:
+        - "get"
+        - "list"
+        - "patch"
+      apiGroups:
+        - ""
+      resources:
+        - "pods"
```
wanghaoran1988, line 23: same here.

scripts/CSSRE/rolling-restart-brokers/metadata.yaml
pat-cremin: I'd assumed from testing it needed clusterRoles, as other CSSRE examples were using cluster-wide roles.  I can see the example in SREP and will test and then switch it.

pat-cremin: Actually, there is no fuzzy logic on the namespace limitation, so it would need to remain as  clusterRoleRules, as the namespace is unknown and requested as part of the script. unless there is a way to feed the provided namespace into the metadata after the fact, I can't see how I could specify the NS.

scripts/CSSRE/rolling-restart-brokers/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Usage
+
+### Running locally
+
+```bash
+Usage: ./rolling-restart-brokers.sh -n KAFKA_NAMESPACE [-s STATEFUL_SET | -p POD]
+
+# Examples
+./rolling-restart-brokers.sh -n kafka-ID
```
robshelly, line 13: Just an minor comment here: `Kafka-ID` is used here in place of `KAFKA_NAMESPACE`. Might be best to change to avoid confusion this as kafka ID and kafka namespace are not the same.
`KAFKA_NAMESPACE=kafka-$KAFKA_ID`

robshelly (commented):
Tested out locally and everything works. Just one minor comment on the examples usage in the readme.

One thing thing to note though. I recently became aware of a change coming soon where statefulsets will be migrated to `strimzipodsets`. In anticipation of this work all managed scripts will need to be refactored to be able to handle both statefulsets and strimzipodsets. Might be best to make that change to this script now.
More info: https://issues.redhat.com/browse/MGDSTRM-9317

scripts/CSSRE/rolling-restart-brokers/README.md
pat-cremin: Changed as suggested.

robshelly (approved):
:+1: 
Tested locally. Works with pods, statefulsets and strimzipodsets.

pat-cremin:
@wanghaoran1988 can I get a review on this PR please?

PR #80, opened 2022-08-09
Get transaction info in get-kafka-instance-state.sh

grdryn:


scripts/CEE/get-kafka-instance-state/metadata.yaml
```diff
@@ -83,4 +83,10 @@ envs:
   - key: "since_time"
     description: "This is being passed to use the --since-time flag used by oc adm inspect. Timestamp is in RFC 3339 format. Only one of since-time/since may be used"
     optional: true
+  - key: "kafka_max_transaction_timeout"
+    description: "Max transaction timeout in minutes to use when querying for hanging transactions"
+    optional: true
+  - key: "kafka_num_brokers"
```
k-wall, line 89: is there a use-case for running the command against fewer than the number of brokers the instance has?  If no, I think the script should work out the number of brokers that exist, and run it against all.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -42,6 +42,22 @@ function main() {
   oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-acls.sh \
     --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-acls.txt 2>&1
 
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-transactions.sh \
+    --bootstrap-server localhost:9096 list > "${inspectDir}"/kafka-transactions-list.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-transactions.sh \
```
k-wall, line 48: is this a dup with line 45?

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -42,6 +42,22 @@ function main() {
...
+    --bootstrap-server localhost:9096 list > "${inspectDir}"/kafka-transactions-list.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh -c \
+    "awk 'NR > 1 { print \$1 }' ${inspectDir}/kafka-transactions-list.txt 2>&1 | xargs -L 1 ./bin/kafka-transactions.sh --bootstrap-server localhost:9096 describe --transactional-id \$@ > ${inspectDir}/kafka-transactions-describe.txt 2>&1"
```
k-wall, line 52: Does `${inspectDir}/kafka-transactions-describe.txt` filename need the transaction-id interpolated into it to make it unique.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -42,6 +42,22 @@ function main() {
...
+
+  for b in $(seq 0 "${kafka_num_brokers:-3}"); do
+    oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-transactions.sh \
```
SamBarker, line 55: Is there any reason to do this sequentially? It would be faster to run each  broker in the background.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -42,6 +42,22 @@ function main() {
   oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-acls.sh \
     --bootstrap-server localhost:9096 --list > "${inspectDir}"/kafka-acls.txt 2>&1
 
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-transactions.sh \
+    --bootstrap-server localhost:9096 list > "${inspectDir}"/kafka-transactions-list.txt 2>&1
```
SamBarker, line 46: Would it be worth gathering all transaction related files into a specific sub directory within `inspectDir`

scripts/CEE/get-kafka-instance-state/metadata.yaml
grdryn: > is there a use-case for running the command against fewer than the number of brokers the instance has? If no, I think the script should work out the number of brokers that exist, and run it against all.

Good idea, thanks. I'll try that.

> Also I notice that the [--broker-id BROKER_ID] is marked as optional, do we actually need to pass it at all?

The `find-hanging` command requires either `--topic` or `--broker-id` to limit the scope of the search. It seems easier to pass the broker id.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
```diff
@@ -42,6 +42,22 @@ function main() {
...
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh /opt/kafka/bin/kafka-transactions.sh \
+    --bootstrap-server localhost:9096 list > "${inspectDir}"/kafka-transactions-list.txt 2>&1
+
+  oc exec -n "${resource_namespace}" "${kafkaStatefulSet}" -c kafka -- sh -c \
```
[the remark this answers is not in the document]
grdryn: > I think your intent is to iterate across all the transaction id in the kafka-transactions-list.txt and run a describe on each, right? There's a problem in that. if the kafka instance has for whatever reason accumulated lots of transactions, get-managed-instance-state might take an excessive length of time to run. I suggest an extra parameter that limits the number of transaction describes that will be performed. Default this to something like 25.

Hmm, I hadn't thought of that as a possible problem. :thinking: I'm leaning more towards not running the `describe` on any transactions than just choosing a limited number for now. If we do need that information (I'm not sure that we currently do), then there's no way of knowing if the transaction(s) we actually do want it for would be returned as part of the limited set of 25 results, and we wouldn't really have a way to know what number we have to set to get the results we do want. Maybe we need a separate managed script that just wraps that `describe` sub-command. This is what I was thinking about as a possibility for the `abort` command also. WDYT?

> Separately, your scripting looks suspicious. kafka-transactions-list.txt is on the client, but you are passing in within the shell command executed within the pod. It won't be able to access the file.

Yes, I started testing this after pushing the WIP PR and noticed that too :)

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
grdryn: Either that, or each invocation should append to the file rather than overwriting it.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
grdryn: I guess that could be done, we'd just need to await them before tar-streaming the directory into the logs.

scripts/CEE/get-kafka-instance-state/get-kafka-instance-state.sh
grdryn: Maybe :thinking: I'll consider if that's worth it after I (maybe) move the `describe` stuff out of here to a separate script.

grdryn: Done, thanks for the suggestion!

grdryn:
I've updated this to pull the describe part out into a separate script that can be called on a specific transaction, and addressed the remaining comments. @k-wall @SamBarker Please take another look if you have a moment, and let me know what you think of this approach. Thanks!

(I don't know what's going on with that failing build...I don't have access to see the logs :disappointed:)

grdryn:
@SamBarker could you take a look at this when you have a moment, to see if you're happy with it now? Thanks!

SamBarker (approved):
LGTM

grdryn:
@wanghaoran1988, @MitaliBhalla could you please take a look at this and approve if you have no objections? Thanks a lot! :slightly_smiling_face:

grdryn:
/hold

I think I need to make another change here

grdryn:
/hold cancel

I've updated this to be fine again. I just realised while doing some extra testing that I had specified `wait -f`, and in the version of Bash in the container image, `wait` doesn't yet have `-f`, so `wait` is enough.

scripts/CEE/describe-kafka-transaction/metadata.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - CSSRE
+  - CEE
+rbac:
+  clusterRoleRules:
+    - verbs:
+        - "create"
+      apiGroups:
+        - ""
+      resources:
+        - "pods/exec"
```
wanghaoran1988, line 17: Do you need this permission at cluster scope?

grdryn: @wanghaoran1988 I _think_ so, but I'm happy to hear of any suggestions that you have. My understanding is that the script will run in the openshift-managed-scripts namespace, but for this script we pass a parameter called `resource_namespace` at run time to indicate the namespace in which the pods we need to exec to. So unless there's some magic that I don't yet understand (can we somehow provide it for `resource_namespace` dynamically?), we need to give this permission at the cluster scope so that it will have that permission in the `resource_namespace` namespace. WDYT?

PR #82, opened 2022-08-17
Change oc exec commands to use services instead of statefulsets

racheljpg:
This is a draft pull request to get feedback and test some changes.
Changing the oc exec commands to use services instead of statefulsets. Strimzi is moving away from statefulsets, so we want this script to be able to support both approaches.


grdryn (approved):
/lgtm

I have reviewed these changes and tried them out locally.

PR #83, opened 2022-08-24
Update scripts to use services instead of statefulsets to support strimzipodsets

racheljpg:
Making this as a draft PR to ask some questions on the work I'm doing to refactor these scripts to use services instead of statefulsets, for the work on migrating to using strimzipodsets :)

### DIFF
--- a/scripts/CEE/describe-kafka-transaction/metadata.yaml
+++ b/scripts/CEE/describe-kafka-transaction/metadata.yaml
@@ -19,9 +19,9 @@ rbac:
         - "get"
         - "list"
       apiGroups:
-        - "apps"
+        - ""
       resources:
-        - "statefulsets"
+        - "services"
     - verbs:
         - "get"
       apiGroups:
--- a/scripts/CSSRE/lib/py/exceptions.py
+++ b/scripts/CSSRE/lib/py/exceptions.py
@@ -8,9 +8,9 @@ class MissingEnvironmentVariable(Exception):
         super().__init__(msg)
 
 
-class StatefulsetExecConnectionError(Exception):
+class ServiceExecConnectionError(Exception):
     def __init__(self, msg):
-        details = "This can happen if there are no pods in the statefulset in a ready state or a pod restarted during the process. Ensure there are ready pods and try again."
+        details = "This most likely can happen if there are no pods reached by the service in a ready state or a pod restarted during the process. Ensure there are ready pods and try again. This can also happen if the bootstrap server for the service does not exist/cannot be reached."
         super().__init__(msg, details)
 
 class NotManagedKafkaNamespace(Exception):
--- a/scripts/CSSRE/lib/py/kafka.py
+++ b/scripts/CSSRE/lib/py/kafka.py
@@ -1,5 +1,5 @@
 from py.logger import Logger
-from py.exceptions import StatefulsetExecConnectionError, NotManagedKafkaNamespace
+from py.exceptions import NotManagedKafkaNamespace, ServiceExecConnectionError
 import copy
 
 class Kafka:
@@ -88,7 +88,7 @@ class Kafka:
     """
     with self._oc.project(namespace), self._oc.timeout(self.settings.OC_TIMEOUT_5_MINUTES):
 
-      cmd = "-i statefulset/"+kafka+"-kafka -c kafka -- env - bin/kafka-topics.sh --bootstrap-server localhost:9096 --describe "
+      cmd = "-i svc/"+kafka+"-kafka-bootstrap -c kafka -- env - bin/kafka-topics.sh --bootstrap-server localhost:9096 --describe "
       if topic:
         cmd = cmd + " --topic " + topic
       if filter:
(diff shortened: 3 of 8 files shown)


scripts/CSSRE/lib/py/kafka.py
```diff
@@ -81,23 +81,24 @@ def run_kafka_topics_script(self, namespace, kafka, topic="", filter="") :
     Parameters:
       namespace (str): Namespace of the kafka.
       kafka (str): Name of the kafka cluster.
+      service (str): Name of the service being used to run oc exec against StatefulSet or StrimziPodSet.
```
racheljpg, line 84: Where do I add the line of code that actually sets this? Do I need to write something like above at line 5?

scripts/CSSRE/lib/py/kafka.py
```diff
@@ -81,23 +81,24 @@ def run_kafka_topics_script(self, namespace, kafka, topic="", filter="") :
...
         cmd = cmd + " --topic " + topic
       if filter:
         cmd = cmd + " --" + filter
       try:
         return self._oc.invoke('exec', cmd.split())
       except self._oc.OpenShiftPythonException:
-        raise StatefulsetExecConnectionError(
-          f"Connection to Statefulset failed: statefulset/{kafka}-kafka"
+        raise ServiceExecConnectionError(
+          f"Connection to Service failed: {service}" #service name
```
racheljpg, line 101: Is this giving enough information?

robshelly: Extra information is provided by the custom exception defined in https://github.com/openshift/managed-scripts/blob/main/scripts/CSSRE/lib/py/exceptions.py

Here you're replacing the `StatefulsetExecConnectionError` with a new one, `ServiceExecConnectionError`. You'll need to add the new custom exception to the above file (or just modify `StatefulsetExecConnectionError` if it's not being used anymore). You can provide extra information about what this exception means there.

racheljpg: @robshelly Hi Rob, thank you. I just pushed some new changes and I think I've changed it to what you're saying here, would you mind taking a look when you get the chance? :) Thanks!

scripts/CSSRE/lib/py/exceptions.py
```diff
@@ -8,12 +8,12 @@ def __init__(self, msg):
         super().__init__(msg)
 
 
-class StatefulsetExecConnectionError(Exception):
+class ServiceExecConnectionError(Exception):
     def __init__(self, msg):
-        details = "This can happen if there are no pods in the statefulset in a ready state or a pod restarted during the process. Ensure there are ready pods and try again."
+        details = "This can happen if the bootstrap server for the service does not exist/cannot be reached."
```
robshelly, line 13: The error message about this potentially failing if there aren't pods in a ready state is still valid here. 
In fact I think it's the more important part to mention (and more likely cause of a connection error) is pods crashing. When testing the script I've encountered connection errors because a pod restarts so it's good to note that once a pods i running you can just rerun the script

robshelly (commented):
Looks good. Tested the scripts and rbac locally and everything works. Just one minor change suggested.

Also one favour to ask. I noticed when testing these changes that I named the logger for the get-kafka-topics script wrong. 
https://github.com/openshift/managed-scripts/blob/main/scripts/CSSRE/get-kafka-topics/get-kafka-topics.py#L13
Could you change it to `Kafka Topics` in this PR please? Making changes to managed-scripts and getting them rolled out to prod can sometimes be somewhat of a protracted process. Including that change here would avoid having to go through that process for a typo fix.

scripts/CSSRE/lib/py/exceptions.py
racheljpg: Hey Rob, thank you for testing this. No problem, I've changed the logger name and updated the error message to give some more info on the pods. 
The process for rolling out changes to production is [this](https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts#how-to-promote-to-production), right? Would you like me to start that as soon as this gets merged?

scripts/CSSRE/under-min-isr-partitions/metadata.yaml
```diff
@@ -33,6 +33,12 @@ rbac:
         - "kafka.strimzi.io"
       resources:
         - "kafkas"
+    - verbs:
+        - "list"
+      apiGroups:
+        - "core.strimzi.io"
+      resources:
+        - "strimzipodsets"
```
k-wall, line 41: Given the proposed change to scripts/CSSRE/lib/py/kafka.py switches from using statefulsets to service, why do are RBAC permissions for`strimzipodsets` required?

racheljpg: Hello Keith, thank you! That does make more sense. 
I've updated the metadata files for the script in this pr to have permissions for services and removed statefulset permissions. I also changed the permissions on the [changes I made recently to the gather-kafka-instance-state script](https://github.com/openshift/managed-scripts/pull/82), because I had done the same thing there. I've tested it locally and it seems to still work as expected.

scripts/CEE/describe-kafka-transaction/metadata.yaml
```diff
@@ -19,9 +19,9 @@ rbac:
         - "get"
         - "list"
       apiGroups:
-        - "apps"
+        - "kafka.strimzi.io"
```
k-wall, line 22: services is part of the core api group of kube.  it is not part of strimzi. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#service-v1-core

k-wall: Change looks good to me

k-wall (approved):

Issue #84, opened 2022-08-29
Injection attack possibility in kafka.py.


k-wall:
Currently this library doesn't guard against the injection attacks.  For instance, a malicious user could pass in a specially crafted kafka instance name or topic name (containing a white space) to cause undesired execution effects.  Best practice tells us we should always code to avoid this possibility.

Building the cmd arrays explicitly, rather than splitting on whitespace, would be one possible resolution.

https://github.com/openshift/managed-scripts/blob/3211e575e9d718cb58c11d701fceb3dee26f5fe3/scripts/CSSRE/lib/py/kafka.py#L91

@robshelly

feichashao:
Thank you for raising the issue! Given the PR is merged, are we good to close this issue?

robshelly:
Yes this can be closed now. Thanks for pointing out the issue @k-wall.

PR #85, opened 2022-09-12
Prevent potential injection attack

pbabic-redhat:
There is a potential for an injection attack in the `kafka.py` script. Using a list of commands instead of `cmd.split()` will fix it. 
The issue was raised here: https://github.com/openshift/managed-scripts/issues/84
JIRA: [CSSRE-2200](https://issues.redhat.com/browse/CSSRE-2200)
I tested the changes by creating strimzi and kafka instances locally on CRC and running the script - no issues encountered.

### DIFF
--- a/scripts/CSSRE/lib/py/kafka.py
+++ b/scripts/CSSRE/lib/py/kafka.py
@@ -92,7 +92,7 @@ class Kafka:
       if topic:
         cmd += ["--topic", topic]
       if filter:
-        cmd += ["--", filter]
+        cmd.append(f"--{filter}")
       try:
         return self._oc.invoke('exec', cmd)
       except self._oc.OpenShiftPythonException:
--- a/scripts/CSSRE/lib/py/kafka.py
+++ b/scripts/CSSRE/lib/py/kafka.py
@@ -74,7 +74,7 @@ class Kafka:
     return (num_pods_ready, len(pods))
 
 
-  def run_kafka_topics_script(self, namespace, kafka, topic="", filter="") :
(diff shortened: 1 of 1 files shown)


scripts/CSSRE/lib/py/kafka.py
```diff
@@ -87,14 +87,14 @@ def run_kafka_topics_script(self, namespace, kafka, topic="", filter="") :
...
-
-      cmd = "-i svc/"+kafka+"-kafka-bootstrap -c kafka -- env - bin/kafka-topics.sh --bootstrap-server localhost:9096 --describe "
+      cmd = ["-i", "svc/" + kafka + "-kafka-bootstrap", "-c", "kafka", "--", "env", "-",
+             "bin/kafka-topics.sh", "--bootstrap-server", "localhost:9096", "--describe"]
       if topic:
-        cmd = cmd + " --topic " + topic
+        cmd += ["--topic", topic]
       if filter:
-        cmd = cmd + " --" + filter
+        cmd += ["--", filter]
```
robshelly, line 95: The `--` and `filter` will need to be appended as a single string. This way includes a space between them resulting in the script not picking up the script.

robshelly (changes requested):

scripts/CSSRE/lib/py/kafka.py
pbabic-redhat: Thanks for pointing that out, fixed now!

robshelly (approved):
Test locally and everything works.

PR #87, opened 2022-09-23
Add a standard bash template to managed scripts

supreeth7:
This PR adds a standard bash template to the managed scripts repository.
<hr>

Resolves: [OSD-12093](https://issues.redhat.com/browse/OSD-12093)

### DIFF
--- a/template/README.md
+++ b/template/README.md
@@ -1,0 +1,22 @@
+This guide provides in broad strokes a set of Bash best practices that are expected to be adhered to.
+
+### Best practices and tips
+
+- Include a one-line comment description at the top of the script explaining what the script does.
+- Always deference variables with double quotes. Ex: `"$variable"`
+- Declare all global variables as `readonly`.
+- Global variables always have UPPER_CASE naming.
+- Always use local when setting variables, unless there is reason to use declare.
+- All local variables should have lowercase naming.
+- Always have a `main()` function for runnable scripts, called with `main` or `main "$@"`.
+- Modularize code into functions as much as possible. All code goes into functions.
+- Always use `set -euo pipefail`, fail fast and be aware of exit codes.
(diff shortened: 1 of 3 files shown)


supreeth7:
@feichashao can you please give this a look? Thanks!

/assign @feichashao

template/README.md
```diff
@@ -0,0 +1,22 @@
+This guide provides in the broad strokes a set of bash best practices that are expected to be adhered to.
+
+### Best practices and tips
+
+- Include a one-line comment description at the top of the script explaining what the script does.
+- Always double quote variables.
```
feichashao, line 6: Does the mean when de-referencing, always use double quotes?  eg, "$variable".
This may not apply if using that in a for loop.
list="a b c"
for i in "$list" //<-- the ShellCheck will prompt an error, because the for loop may take it as one single item.

supreeth7: Ah right, might need to reword that to something like:
```suggestion
- Always deference variables with double quotes. Ex: `"$variable"`
```

supreeth7: It is expected that whoever uses managed scripts know not to dereference the iterator var over loops :)

template/README.md
```diff
@@ -0,0 +1,22 @@
+This guide provides in broad strokes a set of bash best practices that are expected to be adhered to.
```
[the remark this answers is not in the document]
supreeth7: {programming/scripting lang} bash practices is the standard.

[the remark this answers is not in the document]
MitaliBhalla: Ack.

supreeth7:
@feichashao a quick test in stage and turns out managed-scripts does not support libraries (yet).
WIll revert back to the standard bash template.

/hold

supreeth7:
@feichashao this is now ready to be re-reviewed. 🙇🏽 

/unhold

PR #94, opened 2022-11-08
[MGDCTRS-1661] Collecting oc events in text format in addition to yaml format. Makes…

tmielke:
… it easier for humans to read.


supreeth7:
Fine by me...

/lgtm

PR #104, opened 2023-03-09
OSD-15149 - Creating a new describe-nodes backplane scripts

squirrd:
The new script allows any backplane user to gain access to the lost "cluster node resource usage" which was lost durring the RBAC tightening that occurred late last year.

It pretty much runs `oc describe nodes` but provides validation and a few nice arguments to help with choosing specific nodes.

### DIFF
--- a/scripts/SREP/describe-nodes/README.md
+++ b/scripts/SREP/describe-nodes/README.md
@@ -1,0 +1,44 @@
+# describe-nodes script
+
+## Purpose
+
+Provide the full functionality of `oc describe nodes`, as the standard `oc describe nodes` in backplane restricts access to customer workload information.
+It supports 4 different modes for listing the nodes: 
+* requesting all nodes
+* Requesting nodes from a list
+* Requesting types of nodes (master, infra, worker)
+* Requesting nodes that match a selector
+
+## Parameters and usage 
+### Script usage
+
+```bash
+usage: describe-nodes [--all | --master | --infra | --worker | --nodes <node>,<node>,...]
+  --all      : Describe all nodes in the cluster
+  --master   : Describe the master nodes in the cluster
+  --infra    : Describe the infra nodes in the cluster
+  --worker   : Describe the worker nodes in the cluster
+  --selector : A Label selector to pass to oc describe nodes
+  --nodes    : Describe the listed nodes in the cluster.  The list comprises of the node name separated by a comma with no spaces
+  --help
+
+Argument precedence as only one mode can be used at a time: (the first available is used)
+  --nodes
+  --selector
+  --all
+  --master and|or --worker --and|or infra
+```
+
+### Usage with managed-scripts
+
+For usage with managed-scripts, the options need to be passed through the `SCRIPT_PARAMETERS` environment variable. Here are some examples : 
+
+```bash
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--all"
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--master --infra"
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--nodes ip-10-0-137-48.us-east-2.compute.internal,ip-10-0-135-110.us-east-2.compute.internal" 
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--selector node-role.kubernetes.io=infra"
+```
--- a/scripts/SREP/describe-nodes/README.md
+++ b/scripts/SREP/describe-nodes/README.md
@@ -14,19 +14,22 @@ It supports 4 different modes for listing the nodes:
 
 ```bash
 usage: describe-nodes [--all | --master | --infra | --worker | --nodes <node>,<node>,...]
-  --all      : Describe all nodes in the cluster
-  --master   : Describe the master nodes in the cluster
-  --infra    : Describe the infra nodes in the cluster
-  --worker   : Describe the worker nodes in the cluster
-  --selector : A Label selector to pass to oc describe nodes
-  --nodes    : Describe the listed nodes in the cluster.  The list comprises of the node name separated by a comma with no spaces
-  --help
+  -a, --all             : Describe all nodes in the cluster
+  -m, --master          : Describe the master nodes in the cluster
+  -i, --infra           : Describe the infra nodes in the cluster
+  -w, --worker          : Describe the worker nodes in the cluster
+  -l, --selector        : A Label selector to pass to oc describe nodes
+  -n, --nodes, --node   : Specify the nodes to describe in the cluster separated by a ',' with no spaces
+  -d, --debug           : Enable debugging
+  -h, --help            : Print this help
 
 Argument precedence as only one mode can be used at a time: (the first available is used)
+  --help
   --nodes
   --selector
   --all
   --master and|or --worker --and|or infra
+  The --debug argument can be used at anytime"
 ```
 
 ### Usage with managed-scripts
--- a/scripts/SREP/describe-nodes/script.sh
+++ b/scripts/SREP/describe-nodes/script.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+#
+# Replace a master machine.
+# Input:
+#  (Env) MACHINE
+#    name of the machine resource in cluster
+
+# Exit on error
+set -e
+# Treat unset variables as error
+set -u
+# Print expanded commands with arguments before executing them
+# set -x
+# Set the return value in a set of pipes to the rightmost failed command
+set -o pipefail
+
+## Display error and end
+function error-out() {
+  echo "$0: $1"
+  echo "\
+usage: describe-nodes [--all | --master | --infra | --worker | --nodes <node>,<node>,...]
+  --all      : Describe all nodes in the cluster
+  --master   : Describe the master nodes in the cluster
+  --infra    : Describe the infra nodes in the cluster
+  --worker   : Describe the worker nodes in the cluster
+  --selector : A Label selector to pass to oc describe nodes
+  --nodes    : Describe the listed nodes in the cluster.  The list comprises of the node name seperated by a comma with no spaces  ; 
+  --help     : Print this help
+
+Argument precedence as only one mode can be used at a time: (the first available is used)
+  --nodes
+  --selector
+  --all
+  --master and|or --worker --and|or infra"
+  exit 1
+}
+
+function add2nodes() {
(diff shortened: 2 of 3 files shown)


scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+        || [ -z ${ARG_ARRAY[1]} ] \
+        || [[ ${ARG_ARRAY[1]} =~ ^-.* ]]
+      then
+        error-out "'--selector' must contain a node selector"
+      fi
+      badchars="$(echo ${ARG_ARRAY[1]} | tr -d '[:alnum:]' | tr -d '-' | tr -d '.'  | tr -d '/'  \
+          | tr -d ',' | tr -d ':' | tr -d ' ' | tr -d '(' | tr -d ')' | tr -d '!' | tr -d '=')"
+      if [ ! -z ${badchars} ]
+      then
+        error-out "--selector ilegal caharcters in selector.  These characters '${badchars}' are ileagal"
```
mrbarge, line 136: Just being pedantic
```suggestion
        error-out "--selector illegal characters in selector.  These characters '${badchars}' are illegal"
```

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+  esac
+done
+
+# If --nodes is supplied use that and ignore all other arguments
+# else if --selector is supplied use that ignore the otheres arguments
+# else if --all is supplied use that ignore the otheres arguments
+# else if any of the --master or --infra or --worker arguments are supplied use them all
+if [ ! -z "$nodes" ]
+then
+
```
mrbarge, line 156: No actual command between the `then` and the `elif` is invalid bash - shellcheck (in the PR check) is currently picking this up as a problem. If nothing is meant to happen here, that whole if test can be removed.

Note that as soon as that's fixed, shellcheck is going to give present you with a handful of other issues which need to be addressed to pass linting checks. I'll cover a few of them here, but there's a bunch, so it'll probably be easier to review the PR check output (or run `shellcheck` manually over your script) to fix them all.

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+  fi
+  if [ "$worker" = true ]
+  then
+    add2nodes $(list-selector "node-role.kubernetes.io!=infra,node-role.kubernetes.io/worker")
+  fi
+fi
+
+# If he nodes variable is empty, it means the no arguments were supplied on the command line
+# Or the selector(s) did not pick any nodes.  This is an error
+if [ ! -n "$nodes" ]
```
mrbarge, line 182: Ensure that each variable is string-quoted and throw {} around them

Additionally, `-z` can be used instead of not-minus-n.

```suggestion
if [ -z "${nodes}" ] 
```

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+done
+
+# If --nodes is supplied use that and ignore all other arguments
+# else if --selector is supplied use that ignore the otheres arguments
+# else if --all is supplied use that ignore the otheres arguments
+# else if any of the --master or --infra or --worker arguments are supplied use them all
+if [ ! -z "$nodes" ]
+then
+  
+elif [ ! -z "$selector" ]
```
mrbarge, line 157: Can use `-n` for a non-empty test.

```suggestion
elif [ -n "${selector}" ]
```

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+then
+  nodes=$(list-selector $selector)
+elif [ "$all" = true ]
+then
+  add2nodes $(list-selector "node-role.kubernetes.io/master")
```
mrbarge, line 162: Surround the argument with quotes. (this may look wrong, ie. nested quotes, but it is valid bash)

```suggestion
  add2nodes "$(list-selector "node-role.kubernetes.io/master")"
```

scripts/SREP/describe-nodes/script.sh
squirrd: Fixed and re- tested

`If nothing is meant to happen here, that whole if test can be removed.`  Unfortunately it is required to enforce the precedence order since I only allow one mode of the the command to run...  See the comment at the start of the iff statement.

But it is poor programming, I have fliped the test and moved the ELIF to the true section of the test.

In the previous commit there was a break, but I realised this was also invalid so I had removed it... I was in 2 minds weather to drop a non impacting cmd (equivalent to python `skip`) in there (like `echo -n`) or re-write,..... My 2 minds had walked away from the problem with out solving it.

scripts/SREP/describe-nodes/script.sh
squirrd: Fixed (mostly) - Not completely on board with the quoting and bracketing, but agree with keeping to a standard... Mostly fixed lets see what the linter says about the rest.

scripts/SREP/describe-nodes/script.sh
squirrd: Fixed - all ! -n|-z : Sometimes you just cant see your blind spots until you are poked.

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,189 @@
...
+      error-out "Unknown argument"
+      ;;
+  esac
+done
+
+# If --nodes is supplied use that and ignore all other arguments
+# else if --selector is supplied use that ignore the otheres arguments
+# else if --all is supplied use that ignore the otheres arguments
+# else if any of the --master or --infra or --worker arguments are supplied use them all
+if [ ! -z "$nodes" ]
```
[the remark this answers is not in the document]
squirrd: Fixed

scripts/SREP/describe-nodes/script.sh
squirrd: Leavening for the linter along with my other digressions... I assume it will scold me appropriately as well (-:

scripts/SREP/describe-nodes/README.md
```diff
@@ -0,0 +1,47 @@
...
+For usage with managed-scripts, the options need to be passed through the `SCRIPT_PARAMETERS` environment variable. Here are some examples : 
+
+```bash
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--all"
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--master --infra"
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--nodes ip-10-0-137-48.us-east-2.compute.internal,ip-10-0-135-110.us-east-2.compute.internal" 
+
+ocm backplane managedjob create SREP/describe-node -p SCRIPT_PARAMETERS="--selector node-role.kubernetes.io=infra"
```
mrbarge, line 46:
```suggestion
ocm backplane managedjob create SREP/describe-nodes -p SCRIPT_PARAMETERS="--all"

ocm backplane managedjob create SREP/describe-nodes -p SCRIPT_PARAMETERS="--master --infra"

ocm backplane managedjob create SREP/describe-nodes -p SCRIPT_PARAMETERS="--nodes ip-10-0-137-48.us-east-2.compute.internal,ip-10-0-135-110.us-east-2.compute.internal" 

ocm backplane managedjob create SREP/describe-nodes -p SCRIPT_PARAMETERS="--selector node-role.kubernetes.io=infra"
```

scripts/SREP/describe-nodes/script.sh
```diff
@@ -0,0 +1,186 @@
...
+
+# If the nodes variable is empty, it means the no arguments were supplied on the command line
+# Or the selector(s) did not pick any nodes.  This is an error
+if [ -z "${nodes}" ]  # retest
+then
+  error-out "No nodes selected"
+fi
+
+nodes=$(echo "${nodes}" | tr "," " ")
+oc describe nodes "${nodes}"
```
mrbarge, line 185: Unfortunately surrounding the `${nodes}` with quotes to satisfy shellcheck breaks the script because it treats the space-separated list of nodes as a single argument.

Instead, to do this in a way that makes both the `oc` and `shellcheck` happy, we should turn the `nodes` string into a bash array:

```suggestion
IFS=',' read -r -a nodearray <<< "${nodes}"
oc describe nodes "${nodearray[@]}"
```

(I tested this and it seemed to work)

squirrd: The issue highlighted by SpellChecker is that when the line `oc describe nodes ${nodes}` is unquoted it is prone to abuse by globbing and therefore open to malicious intent from the command line.  (Well not really because I validate the input of the command line) So the recommended approach is to quote the variable

The main issue caused by quoting `oc describe nodes "${nodes}"` the variable is that now the OC CLI only sees 1 argument.  I think the following suggestion should be enough?

Checking exceptions in https://www.shellcheck.net/wiki/SC2086

This may be a better alternative?
```suggestion
nodes=($(echo "${nodes}" | tr "," " "))
oc describe nodes "${nodes[@]}"
```

Similar to this command line POC
```
dsquirre@dsquirre-mac describe-nodes % args="-l,-a,-r,-t"
dsquirre@dsquirre-mac describe-nodes % args=($(echo "${args}" | tr "," " "))
dsquirre@dsquirre-mac describe-nodes % echo ls "${args[@]}"
ls -l -a -r -t
dsquirre@dsquirre-mac describe-nodes % ls "${args[@]}"
total 40
drwxr-xr-x  14 dsquirre  staff   448  3 Mar 13:44 ..
drwxr-xr-x  13 dsquirre  staff   416  7 Mar 15:21 test
-rw-r--r--   1 dsquirre  staff  1007 10 Mar 10:21 metadata.yaml
-rw-r--r--   1 dsquirre  staff  1925 10 Mar 10:23 README.md
drwxr-xr-x   7 dsquirre  staff   224 10 Mar 11:13 .
-rwxr--r--   1 dsquirre  staff   129 10 Mar 14:48 run.sh
-rwxr--r--   1 dsquirre  staff  6132 13 Mar 16:10 script.sh
dsquirre@dsquirre-mac describe-nodes %
```

mrbarge: That suggested solution, whilst working, seems to trigger SC2207 when I run shellcheck over it, so it may be trading one style issue for another. The suggested solution I provided in my comment has no shellcheck issues in case you want to do that instead.

wanghaoran1988:
@squirrd Could you squash the commits?

squirrd:
@wanghaoran1988   Squash Done!   Thank you for looking into this

PR #106, opened 2023-03-21
Library support for backplane managed scripts

reetika-vyas:
Adding library support for backplane managed scripts

Fixes : [OSD-13310](https://issues.redhat.com/browse/OSD-13310)

PR #109, opened 2023-05-05
Managed job library support

hectorakemp:
Rewrite of https://github.com/openshift/managed-scripts/pull/106. 

Adds the ability to `source` a bash library from anywhere within the `scripts` directory.

If a script is in `/scripts/SREP/script.sh`, and the library is also in `/scripts/lib/lib.sh`, it must be sourced like so in `script.sh`:
`source /managed-scripts/lib/lib.sh`. This is due to how the managed scripts are bundled in the `Dockerfile`. 

On staging/production, these will be built in. This behaviour is emulated in backplane test jobs by inlining the library scripts.

### DIFF
--- a/scripts/examples/lib-sourcer/metadata.yaml
+++ b/scripts/examples/lib-sourcer/metadata.yaml
@@ -1,0 +1,9 @@
+file: script.sh
+name: lib-sourcer
+shortDescription: Checks if the directory /foo/bar exists and prints that to the screen.
+description: |
+  A toy example showing how to import bash libraries from the `/libs` directory within a script.
+author: hkemp
+allowedGroups: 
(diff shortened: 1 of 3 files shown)


supreeth7 (commented):
@hectorakemp this looks good overall.

If I'm not wrong, the original ask in https://issues.redhat.com/browse/OSD-13310 was to have a standard library that can be imported throughout the managed scripts repository.  This PR implements a folder-constrained library support, can it be made global?

hectorakemp:
@supreeth7 the example I did may be misleading - it is global. Any file in the `scripts` directory can be sourced by any shell script. Should I update the example to reflect this?

PR #113, opened 2023-07-18
Fix to allow CEE to create secret pcap-collector

rafaelguerra01:
Fix to allow CEE to create secret pcap-collector

### DIFF
--- a/scripts/CEE/pcap-collector/metadata.yaml
+++ b/scripts/CEE/pcap-collector/metadata.yaml
@@ -61,6 +61,7 @@ rbac:
     - verbs:
         - "get"
         - "delete"
+        - "create"
       apiGroups:
         - ""
       resources:


feichashao:
Hi @rafaelguerra01 , thanks for the PR!

I would like more context around this fix.

From the description in `metadata.yaml`, the action to create the secret is
```
$ oc create secret generic pcap-collector-creds --from-literal=username=username_from_step_1 --from-literal=internal=result_from_step_2 --from-literal=password=token_from_step_3  --from-literal=caseid=support_case_number_step_0 -n openshift-backplane-managed-scripts
```

IIUC, the command is not run by the script, but the user to run it manually. So I suppose the fix should be in the RBAC of the users https://github.com/openshift/managed-cluster-config/blob/master/docs/backplane/guideline.md .  The RBAC in metadata.yaml is for the script itself.

I could be wrong as I may be lacking context. Appreciate if you could share more details and we can bring it forward.

cc @hectorakemp

hectorakemp:
That's correct @feichashao - it needs to be in managed-cluster-config - see here (https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/srep/30-cee-pcap-collector.Role.yml), here (https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/cee/60-cee-pcap-collector.RoleBinding.yml), here (https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/csm/60-csm-pcap-collector.RoleBinding.yml), and here ((https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/mobb/60-mobb-pcap-collector.RoleBinding.yml)

hectorakemp:
@feichashao - PR up here: https://github.com/openshift/managed-cluster-config/pull/1756. We can probably close this off now. Thank you very much @rafaelguerra01 for bringing this to our attention - once that pull request that I have linked is merged, it should work as intended.

PR #114, opened 2023-07-20
make the oc-hc bin ready during build phase

bmeng:


Dockerfile
```diff
@@ -29,6 +29,22 @@ RUN export OC_LINUX_X86_CLIENT=$(cat sha256sum.txt | grep openshift-client-linux
 RUN sha256sum --check --ignore-missing sha256sum.txt
 RUN tar --extract --gunzip --no-same-owner --directory /out oc --file *.tar.gz
 
+### Temporary solution to integrate the oc-hc
+ENV OC_HC_TAR_URL="https://github.com/givaldolins/openshift-cluster-health-check/releases/download/v0.1.2/oc-hc-v0.1.2-linux-amd64.tar.gz"
+ENV OC_HC_TAR_MD5="https://github.com/givaldolins/openshift-cluster-health-check/releases/download/v0.1.2/oc-hc-v0.1.2-linux-amd64.tar.gz.md5"
```
feichashao, line 34: For security concern, we might better put the MD5 `12d21a265fe6da4fa9e520e60386f051` here directly instead of downloading.
We can approve the MD5 at this time, so that we know the future download of this binary is the approved one.
The risk of downloading is that the 3rd party repo can also change the md5 file when they inject some unapproved code in the same release.

PR #122, opened 2023-08-21
OSD-17959: Add yq binary into managed-scripts container image

samanthajayasinghe:
## What 
Adding yq binary to the managed script docker image 

## Why 
Search yaml files based on criteria

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,22 @@ RUN md5sum -b oc-hc-v0.1.3-linux-amd64.tar.gz | grep $OC_HC_MD5
 # Extract the binary
 RUN tar xzf oc-hc-v0.1.3-linux-amd64.tar.gz --directory /out
 
+
+### Attach yq binary into the image
+ENV YQ_URL="https://github.com/mikefarah/yq/releases/download/v4.35.1/yq_linux_amd64.tar.gz"
+ENV YQ_MD5="be32a02446da244ebe5b94c3834b2c97"
+RUN mkdir -p /yq
+WORKDIR /yq
+
+# Download the yq release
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,6 +44,21 @@ RUN md5sum -b oc-hc-v0.1.3-linux-amd64.tar.gz | grep $OC_HC_MD5
...
+### Attach yq binary into the image
+ENV YQ_URL="https://github.com/mikefarah/yq/releases/download/v4.35.1/yq_linux_arm64.tar.gz"
+RUN mkdir -p /yq
+WORKDIR /yq
+
+# Download the yq release
+RUN curl -sSLf -O $YQ_URL
+
+# Check md5sum for the downloaded tar
+RUN md5sum yq_linux_arm64.tar.gz  | md5sum -c
```
bmeng, line 57: What do we want for this md5sum?

samanthajayasinghe: Added the static MD5 value and `grep` it

Dockerfile
```diff
@@ -75,6 +91,7 @@ RUN  yum -y install --disableplugin=subscription-manager \
      && yum --disableplugin=subscription-manager clean all
 COPY --from=build-stage0 /out/oc  /usr/local/bin
 COPY --from=build-stage0 /out/oc-hc  /usr/local/bin
+COPY --from=build-stage0 /out/yq_linux_amd64  /usr/local/bin
```
bmeng, line 94: do we want to rename this to `yq`?

samanthajayasinghe: It's not needed, but I renamed it for clean binary name

PR #123, opened 2023-08-23
OSD-17842: Create a managed-script for SREP team to update the upgrade policy state

xiaoyu74:
## What
Create a managed-script for SREP team to update the upgrade policy state

## Why

- We have the above sop for SRE team to force update the upgradepolicy when we face the issue where the upgrade status is different in OCM and the cluster.
- The operation will need SRE to fetch the pull secret on the cluster and patch the upgradepolicy in OCM.
- It would be better if we can use the managed script to implement this to reduce the manual work and also manage the SRE action on reading the secret on the cluster.

## For reference
Jira task: https://issues.redhat.com/browse/OSD-17842

[SOP](https://github.com/openshift/ops-sop/blob/682d31366d694029572d3645cd3d6191448981e6/v4/knowledge_base/set-upgrade-state.md)


scripts/SREP/update-upgrade-policy-state/metadata.yaml
```diff
@@ -0,0 +1,30 @@
+file: script.sh
+name: update-upgrade-policy-state
+description: Update the state of the current cluster's upgrade policy using the desired state.
+author: sbai
+allowedGroups: 
+  - SREP
+  - CSSRE
```
MitaliBhalla, line 7:
```suggestion
  - CSSRE
```
I think CSSRE is LPSRE now?

scripts/SREP/update-upgrade-policy-state/script.sh
```diff
@@ -0,0 +1,101 @@
...
+            DESIRED_STATE="${arg#DESIRED_STATE=}"
+            ;;
+        *)
+            echo "Invalid argument: $arg"
+            exit 1
+            ;;
+    esac
+done
+
+# Validate that a desired state argument is provided
```
MitaliBhalla, line 69:
```suggestion
# Validate that a desired state argument is provided
```
Though , I think it'll be mostly handled already but can we also have another clean clear check on checking whether the DESIRED STATE passed by the user is a valid one?

MitaliBhalla (changes requested):

scripts/SREP/update-upgrade-policy-state/metadata.yaml
MitaliBhalla: Looking at older PRs like these https://github.com/openshift/managed-scripts/pull/105/files , I guess you are right. LPSRE and CSSRE are being considered different.

scripts/SREP/update-upgrade-policy-state/script.sh
supreeth7: Yeah, we can purge all the `CSSRE` references, only `LPSRE` stands true

scripts/SREP/update-upgrade-policy-state/metadata.yaml
supreeth7: Yeah, we can purge all the CSSRE references, only LPSRE stands true

xiaoyu74: So, I assume it's all good for the current setting of the allowed groups. Please let me know if anything I misunderstood.

scripts/SREP/update-upgrade-policy-state/script.sh
xiaoyu74: Thanks for your time to review. It's a good catch. 

Since it's an initial implementation based on my understanding in the SOP,  It's not quite clear mentioning the expected upgrade_policy state we want to forcefully set and any dependencies or restrictions for the current state to be updated to the desired state.

For now, I will define the DESIRED_STATE as one of `started`, `scheduled` or `cancelled` and update the code accordingly.

scripts/SREP/update-upgrade-policy-state/metadata.yaml
MitaliBhalla: We can remove the CSSRE reference and can just stick to LPSRE.

scripts/SREP/update-upgrade-policy-state/script.sh
```diff
@@ -0,0 +1,127 @@
...
+# Parse the arguments passed to the script to determine the desired state.
+parse_arguments() {
+    for arg in "$@"; do
+        case "$arg" in
+            DESIRED_STATE=started)
+                DESIRED_STATE="started"
+                ;;
+            DESIRED_STATE=scheduled)
+                DESIRED_STATE="scheduled"
+                ;;
```
mrbarge, line 63: I'm interested in how we landed on these three states as the ones an SRE can switch to. I don't think there should be any case where SRE should be changing to a `scheduled` state, as that is controlled wholly by OCM. I'm similarly struggling to think of cases where an SRE would/should be moving the state into `started`, are there any?

`completed`, `failed` and `cancelled` are probably the ones I can think are valid transitions we should allow an SRE to manually move to.

bmeng: @mrbarge Do we even need the status `completed` or `failed` here? I think we only give the ability to set `cancelled` should be sufficient.

mrbarge: I'd agree I don't really see much of a use case for `failed`.. however I think `completed` is valid if the cluster has actually upgraded to the version that the upgrade policy says it needs to be on, but MUO somehow isn't doing it.

Worth noting that I'm in the middle of fixing a lot of the need to do this type of handling in MUO with OSD-17840, which is currently in a PR/test phase. But the script is still valuable to have just in case.

mrbarge: That sounds fine to me.

xiaoyu74: @mrbarge @bmeng - Thanks a lot for your review feedback. 
Sorry, I wasn't quite clear about what are the valid desired state we should set in the first place and double checked with Bo. I will implement only with the `cancelled` state at this stage.

scripts/SREP/update-upgrade-policy-state/metadata.yaml
```diff
@@ -0,0 +1,27 @@
+file: script.sh
+name: update-upgrade-policy-state
+description: Cancel the upgrade policy of the current OpenShift cluster.
+author: sbai
+allowedGroups:
```
bmeng, line 5: I do not think we would want this to be executed by the other team than SREP?

scripts/SREP/update-upgrade-policy-state/script.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+# 2. Get the cluster UUID.
+CLUSTER_UUID=$(get_cluster_uuid)
+
+# 3. Get the cluster ID.
+CLUSTER_ID=$(get_cluster_id "${CLUSTER_UUID}")
+
+# 4. Get the policy UUID.
+POLICY_UUID=$(get_policy_uuid "${CLUSTER_ID}")
```
bmeng, line 70: We should better handle the error here if the policy cannot be fetch?

Dockerfile
```diff
@@ -72,6 +72,28 @@ RUN unzip awscliv2.zip
 # Install the bins to the /aws/bin dir so the final image build copy is easier
 RUN ./aws/install -b /aws/bin
 
+## Attach ocm binary into the image
+# Setting ENV variables
+ENV OCM_URL="https://github.com/openshift-online/ocm-cli/releases/download/v0.1.67/ocm-linux-amd64"
+ENV OCM_SHA256_URL="https://github.com/openshift-online/ocm-cli/releases/download/v0.1.67/ocm-linux-amd64.sha256"
```
bmeng, line 78: As the other bin, we should write the sha256 value here instead of get them from github.
And it will be a bit easier when we do the check next step.

xiaoyu74: oh, actually, I  just referred the way used for getting shar256 for `oc` binary in this dockerfile and thought may be a convention? `RUN curl -sSLf ${OC_URL}/sha256sum.txt -o sha256sum.txt`

Yeah, it might be easier to hard code the value. Thanks

scripts/SREP/update-upgrade-policy-state/metadata.yaml
xiaoyu74: If this script will only be used by our SREP, I can remove other teams. Actually, I can see most of the scripts in the SREP folder included CEE as allowedGroup.

scripts/SREP/update-upgrade-policy-state/script.sh
xiaoyu74: good catch, I can update to handle the error. Thanks.

scripts/SREP/update-upgrade-policy-state/metadata.yaml
bmeng: let's keep this for SREP only for now, if other teams will use this as well we can add them later

PR #125, opened 2023-08-25
cordon-uncordon-nodes script

givaldolins:
Reasons for writing this script:
- MCS has received in the past many cases where the customer asks us to cordon/uncordon a worker node, and since we don't have privileges to do it we need to engage SRE via OHSS.
- Additionally, sometimes customers ask us to do that in a specific timeframe, which can not be guaranteed by SRE but would be accommodated in the MCS schedule.

I have tested this script on the staging environment

### DIFF
--- a/scripts/CEE/cordon-uncordon-nodes/README.md
+++ b/scripts/CEE/cordon-uncordon-nodes/README.md
@@ -1,4 +1,4 @@
-# Cluster health check
+# Cordon / Uncordon worker nodes
 
 ## Description
 
--- a/scripts/CEE/cordon-uncordon-nodes/README.md
+++ b/scripts/CEE/cordon-uncordon-nodes/README.md
@@ -1,0 +1,15 @@
+# Cluster health check
+
+## Description
+
+This script cordons/uncordons nodes (workers only)
+
+## Usage
+
+```bash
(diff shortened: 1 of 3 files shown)


scripts/CEE/cordon-uncordon-nodes/script.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+if [[ -z "${ACTION}" ]]; then
+    echo 'Variable ACTION cannot be blank'
+    exit 1
+fi
+
+check_worker(){
+    echo "Checking if \"${WORKER}\" is a worker node..."
+    
+    if (oc get nodes -l node-role.kubernetes.io/worker=,node-role.kubernetes.io/infra!= | grep "${WORKER}") &> /dev/null; then
```
cblecker, line 24:
```suggestion
    if (oc get nodes -l node-role.kubernetes.io/worker=,node-role.kubernetes.io/infra!= -oname | grep "${WORKER}") &> /dev/null; then
```

scripts/CEE/cordon-uncordon-nodes/README.md
```diff
@@ -0,0 +1,15 @@
+# Cluster health check
```
cblecker, line 1: Might need updating?

givaldolins: Good catch. Fixed!

PR #132, opened 2023-09-12
osd-17784: fixing json schema and a metadata.yaml

faldanarh:
After testing the json files with the new 'check-jsonschema' package towards the existing jsonschema, it has been identified some issues with the structure of the jsonschema and also with the metadata.yaml of rolling-restart-brokers script. This commit is to fix both files.

### DIFF
--- a/hack/metadata.schema.json
+++ b/hack/metadata.schema.json
@@ -122,22 +122,19 @@
           "type": "array",
           "items": {
             "type": "string"
-          },
-          "required": true
+          }
         },
         "apiGroups": {
           "type": "array",
           "items": {
             "type": "string"
-          },
-          "required": true
+          }
         },
         "resources": {
           "type": "array",
           "items": {
             "type": "string"
-          },
-          "required": true
+          }
         },
         "resourceNames": {
           "type": "array",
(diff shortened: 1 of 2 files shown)


hack/metadata.schema.json
```diff
@@ -122,46 +122,43 @@
...
-          "required": true
+          }
         }
-      }
+      },
+      "required": [ "namespace", "rules" ]
     }
   }
-}
+}
```
wanghaoran1988, line 164: Could you add the empty line back?

scripts/LPSRE/rolling-restart-brokers/metadata.yaml
```diff
@@ -47,4 +44,4 @@ envs:
   - key: "pod"
     description: "Specific pod to manually roll"
     optional: true
-language: bash
+language: bash
```
wanghaoran1988, line 47: ditto

hack/metadata.schema.json
bmeng: any reason we need an extra blank line in the end?

wanghaoran1988: [that's a common across our repos](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file)

faldanarh: Hi, @wanghaoran1988 .
I have amended the commit.
Could you please check?

scripts/LPSRE/rolling-restart-brokers/metadata.yaml
faldanarh: Here too.

PR #137, opened 2023-09-18
Add SFTP upload function

feichashao:
https://access.redhat.com/articles/5594481

We can allow a managed-script to upload files using the unauthenticated way, so that we don't need to handle any secret in the script.

### DIFF
--- a/scripts/lib/sftp_upload/lib.sh
+++ b/scripts/lib/sftp_upload/lib.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 # Description: provide the sftp_upload function for script to call
+#
+# To import the sftp_upload function,
+# include in your script with
+# source /managed-scripts/lib/sftp_upload/lib.sh
 
 # Fail fast and be aware of exit codes
 set -euo pipefail
--- a/scripts/lib/sftp_upload/lib.sh
+++ b/scripts/lib/sftp_upload/lib.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 
 # Constants
 readonly FTP_HOST="sftp.access.redhat.com"
-readonly SFTP_OPTIONS="-o BatchMode=no -b"
+readonly SFTP_OPTIONS=(-o BatchMode=no -b)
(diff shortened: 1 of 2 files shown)


scripts/lib/sftp_upload/lib.sh
```diff
@@ -0,0 +1,43 @@
...
+# Constants
+readonly FTP_HOST="sftp.access.redhat.com"
+readonly SFTP_OPTIONS=(-o BatchMode=no -b)
+readonly KNOWN_HOST='sftp.access.redhat.com,35.80.245.1 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCFQ3l2YVJ0r4MNzAZmTV2kg7rPi4WPeJNcNubvOVA4WwBV6cRsYFkIqtB1unBzTXoZHd7+adtZTgUrJ2BExImyLUQaLBu3KKo4CgGeiZMo8dfDvE2tIe/GwGtyho57TtwJVVUCljvFBvbz8+D6VunsQ6kNU53t8qCaBQNm61twTkdAHP9IESJbC7wWJqjmhmOMTav1OKQDtLEsSDc4I+s+h41LvUfw1lA7RSl9eR13TK9ySpN/uW5nBq7nUNWW5OBc3UbvpdQpDXvdUDbW0rQ2EEWvLkKubhk+RSeY/lH8peOeHYQ5ARPYfFDpo5KsKDDdKa9DfnK8N8APgtzM0r+l'
+
+## Upload a file to sftp.access.redhat.com using the unauthenticated flow.
+## More about the SFTP server:
+## https://access.redhat.com/articles/5594481
+##
+## Usage: sftp_upload <source-filename> <destination-filename>
```
hectorakemp, line 16: Maybe add a small message about how to source the lib? 
i.e 
```
Include in your script with 
source /managed-scripts/lib/sftp_upload/lib.sh
```

hectorakemp (commented):
Overall, script LGTM. If you've tested+verified that it works to upload files using your updated base image, I think it's good to go :+1:

scripts/lib/sftp_upload/lib.sh
feichashao: Thanks!  Added to the beginning of the lib.sh file.

feichashao:
> If you've tested+verified that it works to upload files using your updated base image
Yes, tested with the updated base image and it can upload a file.

PR #139, opened 2023-09-25
Minor improvements over hyperhift must gather script

Tafhim:
This PR adds some minor improvements to the `hs-must-gather` script.

1. Adds a DUMP_GUEST_CLUSTER parameter to optionally include / exclude guest cluster must gather
2. Separates control plane and data plane logs for ease of consumption
3. Renders the SFTP username in lowercase to make it easy to built the SFTP `get` command for download

### DIFF
--- a/scripts/CEE/hs-must-gather/script.sh
+++ b/scripts/CEE/hs-must-gather/script.sh
@@ -12,6 +12,8 @@ DUMP_DIR="${PWD}/${CLUSTER_ID}"
 TODAY=$(date -u +%Y%m%d)
 TARBALL_NAME="${TODAY}_${CLUSTER_ID}_dump.tar.gz"
 TARBALL_PATH="${DUMP_DIR}/${TARBALL_NAME}"
+HC_DUMP_DIR=""
+CP_DUMP_DIR=""
 
 # Function to run a hypershift dump on the HCP namespace
 collect_must_gather() {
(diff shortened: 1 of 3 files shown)


scripts/CEE/hs-must-gather/script.sh
```diff
@@ -5,13 +5,17 @@ set -o nounset
 set -o pipefail
 
 source /managed-scripts/lib/sftp_upload/lib.sh
+#source $(dirname ${0})/../../lib/sftp_upload/lib.sh
```
bmeng, line 8: what does this comment do?

scripts/CEE/hs-must-gather/script.sh
```diff
@@ -5,13 +5,17 @@ set -o nounset
...
 
 # Define expected values
 HS_BINARY_PATH="/usr/local/bin/hypershift"
+#HS_BINARY_PATH="${HOME}/bin/hypershift"
```
bmeng, line 12: same here, why do we need the comment here?

scripts/CEE/hs-must-gather/script.sh
```diff
@@ -34,8 +36,16 @@ collect_must_gather() {
   # Fetch the HostedCluster name
   hc_name=$(oc get HostedCluster.hypershift.openshift.io -l api.openshift.com/id="${CLUSTER_ID}" -n "${hcp_ns}" -o json | jq -r '.items[0].metadata.name')
 
-  echo "Executing hypershift dump on ${hcp_ns} for cluster ${hc_name}"
-  hypershift dump cluster --dump-guest-cluster --namespace "${hcp_ns}" --name "${hc_name}" --artifact-dir "${DUMP_DIR}" --archive-dump=false
+  HC_DUMP_DIR="hostedcluster-${hc_name}"
+  CP_DUMP_DIR="controlplane-${hc_name}"
+
+  if [ "${DUMP_GUEST_CLUSTER}" = "true" ]; then
```
bmeng, line 42: Is there a way we valid the value here is boolean? If user passes other string here, the script would be failed?

PR #146, opened 2023-10-25
OSD-18414 - Update metadata schema to use clusterVersion

Tafhim:
This PR adds the schema update necessary to use clusterVersion validation feature for managed-scripts.

Resolves: [OSD-18414](https://issues.redhat.com//browse/OSD-18414)

### DIFF
--- a/hack/metadata.schema.json
+++ b/hack/metadata.schema.json
@@ -103,6 +103,15 @@
           }
         }
       }
+    },
+    "clusterVersions": {
+      "type": "array",
+      "description": "List of required cluster versions, supporting 'major.minor.patch' format and wildcards",
+      "additionalProperties": false,
+      "items": {
+        "type": "string",
+        "pattern": "^\\d+\\.\\d+\\.\\d+$|^\\d+\\.\\d+\\.\\*$"
+      }
     }
   },
   "additionalProperties": false,


hack/metadata.schema.json
```diff
@@ -103,6 +103,14 @@
           }
         }
       }
+    },
+    "clusterVersions": {
+      "type": "array",
+      "description": "The cluster versions required to execute the script",
```
xiaoyu74, line 109: I know we should have a short description but is that better the description can reflect the actual versioning scenarios something like: 
```
"description": "List of required cluster versions, supporting 'major.minor.patch' format and wildcards",
```

hack/metadata.schema.json
```diff
@@ -103,6 +103,14 @@
...
+      "additionalProperties": false,
+      "items": {
+        "type": "string"
```
xiaoyu74, line 112: Also, do you think we should add a pattern field under the type like:

xiaoyu74: I think it's just a nit which is relevant to this ticket, and I posted the comment 2 weeks ago. Any reason you want to have another ticket for that or any concerns or questions about the pattern? 

I don't have approval permission, if you want to have the current code merged and closeed in this sprint, please ask Bo to approve and I am all good.

PR #147, opened 2023-10-30
OSD-19295: Managed Script to Re-run jobs that fail in openshift-sre-pruning namespace

Dee-6777:
This script is for the alert Pruning Cronjob Error. Here it will detect and delete the failed jobs that are running in the `openshift-sre-pruning` namespace and causing this error, deleting will help recreate new jobs and possibly resolve the alert, and if not SREs can start investigating and find the root cause.

### DIFF
--- a/scripts/SREP/retry-failed-pruning-cronjob/README.md
+++ b/scripts/SREP/retry-failed-pruning-cronjob/README.md
@@ -1,0 +1,13 @@
+# RETRY PRUNING CRON JOB
+
+## Purpose
+
+This script will help re-run the jobs which are failed in `openshift-sre-pruning` namespace
+
+## Usage
+This script will detect and delete the failed jobs that are running in the `openshift-sre-pruning` namespace and causing pruning-cron-job error, deleting will help recreate new jobs and possibly resolve the alert, and if not SREs can investigate and find the root cause.
+
+## Create the ManagedJob to restart the failed jobs
+```
+ocm backplane managedjob create SREP/fix-pruning-cronjob 
+```
--- a/scripts/SREP/retry-failed-pruning-cronjob/metadata.yaml
+++ b/scripts/SREP/retry-failed-pruning-cronjob/metadata.yaml
@@ -1,0 +1,26 @@
+file: script.sh
+name: fix-pruning-cron-job-error
+description: Retry the failed jobs in openshift-sre-pruning namespace
+author: Dee-6777
+allowedGroups:
+  - SREP
+labels:
+  - key: OSD_TYPES
+    description: Compatible cluster types for this script
+    values:
+      - OSD
+rbac:
+    roles:
+      - namespace: "openshift-sre-pruning"
+        rules:
+          - verbs:
+            - "get"
+            - "list"
+            - "watch"
+            - "delete"
+            - "create"
+            apiGroups:
+            - "batch"
+            resources:
+            - "jobs"
+language: bash
--- a/scripts/SREP/retry-failed-pruning-cronjob/script.sh
+++ b/scripts/SREP/retry-failed-pruning-cronjob/script.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+# Delete and recreate jobs that fail in openshift-sre-pruning namespace and producing prunining cron job error.
+
+set -x
+set -e
+set -o nounset
+set -o pipefail
+
+readonly PRUNING_NS=openshift-sre-pruning
+
+BUILDS_PRUNER=()
+DEPLOYMENTS_PRUNER=()
+FAILED_JOBS=()
+BUILDS_PRUNER_FAILED=()
+DEPLOYMENTS_PRUNER_FAILED=()
+
+BUILDS_PRUNER+=($(oc get jobs -n "$PRUNING_NS" -o=jsonpath='{.items[?(@.metadata.ownerReferences[*].name=="builds-pruner")].metadata.name}'))
+DEPLOYMENTS_PRUNER+=($(oc get jobs -n "$PRUNING_NS" -o=jsonpath='{.items[?(@.metadata.ownerReferences[*].name=="deployments-pruner")].metadata.name}'))
+
+detect_job() {
+  echo "INFO: finding jobs producing error"
+  FAILED_JOBS+=($(oc get jobs -n "$PRUNING_NS" -o=jsonpath='{.items[?(@.status.failed>1)].metadata.name}'))
+}
+
+delete_job() {
+  if [[ ${#FAILED_JOBS[*]} == 0 ]]; then
+    echo "INFO: no failed jobs found exiting.."
+    return 1
+  else
+    echo "INFO: deleting jobs producing error"
+    for jobs in "${FAILED_JOBS[@]}"; do
+      oc delete job "$jobs" -n "$PRUNING_NS"
+    done
+  fi
+}
+
+rerun_job() {
+  for jobs in "${FAILED_JOBS[@]}"; do
+    if [[ ${BUILDS_PRUNER[@]} =~ ${jobs} ]]; then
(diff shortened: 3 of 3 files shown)


scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,46 @@
...
+
+readonly PRUNING_NS=openshift-sre-pruning
+FailedJobs=()
+
+detect_job() {
+  echo "INFO: finding jobs producing error"
+  for current_ns in ${PRUNING_NS}; do
+    readarray -t Jobs < <(oc get job -n "$current_ns" | awk '{print $1}') 
+    Jobs=("${Jobs[@]:1}")
+    for job in "${Jobs[@]}"; do
```
bmeng, line 17: Read the jobs in a loop should work, but it is kind of less efficiency, is it possible we use the oc built-in function to get the failed jobs directly?

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,46 @@
...
+    return 1
+  else
+    echo "INFO: deleting jobs producing error"
+    for jobs in "${FailedJobs[@]}"; do
+      oc delete job "$jobs" -n "$PRUNING_NS"
+    done
+  fi
+}
+
+main() {
```
bmeng, line 39: We should retry the same failed job after deleted here.

scripts/SREP/retry-pruning-cronjob/metadata.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - verbs:
+            - "get"
+            - "list"
+            - "watch"
+            - "delete"
+            apiGroups:
+            - "batch"
+            resources:
+            - "jobs"
+    clusterRoleRules:
```
bmeng, line 25: I am not sure if the clusterRoleRules are needed here.

Dee-6777: Thanks, I've removed that. We don't need permission for any cluster-scoped resources or across all namespaces.

scripts/SREP/retry-pruning-cronjob/script.sh
bmeng: yes, once would be enough, if retry fails, that means there are two failures in total, it should not be an intermittent issue in that case.

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,45 @@
...
+
+delete_job() {
+  if [[ ${#FailedJobs[@]} == 0 ]]; then
+    echo "INFO: no failed jobs found exiting.."
+    return 1
+  else
+    echo "INFO: deleting jobs producing error"
+    for jobs in "${FailedJobs[@]}"; do
+      oc delete job "$jobs" -n "$PRUNING_NS"
+      FailedJobs=("${FailedJobs[@]:1}")
```
xiaoyu74, line 27: Any reason you add this `FailedJobs=("${FailedJobs[@]:1}")`?  

The current logic of the `delete_job` function seems to be to delete each job in the FailedJobs array and then remove it from the array. 

However, `FailedJobs=("${FailedJobs[@]:1}")` is not necessary in this specific script, because the FailedJobs array is not used after the delete_job function. Once all jobs are deleted, the script exits.

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,45 @@
...
+    done
+  fi
+}
+
+retry_job() {
+  detect_job
```
xiaoyu74, line 33: This function looks redundant since it is doing the same thing as the main function.

From the main() function: 
```
main() {
  detect_job
  delete_job
  retry_job
  echo "Succeed."
  exit 0
}
```
From the execution order above, looks you detect and delete failed jobs, then call `retry_job`, which again detects and deletes failed jobs. However, there is **no actual retrying of the jobs happening**,  you are just deleting them?

xiaoyu74, line 33: This function is redundant since it is doing the same thing as the main function. 

If it's our intention to retry (actually, it's just detect and delete) one more time, we just have the `retry_job` to be called twice in the main, may be just renaming the function to better reflect it's actual behavior? 
```
handle_failed_jobs() {
  detect_job
  delete_job
}

main() {
  handle_failed_jobs
  handle_failed_jobs
  echo "Succeed."
  exit 0
}
```

Dee-6777: Hi @xiaoyu74 , as discussed with meng bo, it was actually intended to create the failed job. I've done that but waiting for further reviews from all.

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,53 @@
...
+set -x
+set -e
+set -o nounset
+set -o pipefail
+
+readonly PRUNING_NS=openshift-sre-pruning
+
+Builds_Pruner=()
+Deployements_Pruner=()
+FailedJobs=()
```
bmeng, line 13: Can we set the variable as all upper cases? like `FAILED_JOBS`

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,53 @@
...
+
+Builds_Pruner+=($(oc get jobs -n "$PRUNING_NS" -o=jsonpath='{.items[?(@.metadata.ownerReferences[*].name=="builds-pruner")].metadata.name}'))
+Deployements_Pruner+=($(oc get jobs -n "$PRUNING_NS" -o=jsonpath='{.items[?(@.metadata.ownerReferences[*].name=="deployments-pruner")].metadata.name}'))
```
bmeng, line 16: This is not being used?

scripts/SREP/retry-pruning-cronjob/script.sh
```diff
@@ -0,0 +1,53 @@
...
+  else
+    echo "INFO: deleting jobs producing error"
+    for jobs in "${FailedJobs[@]}"; do
+      oc delete job "$jobs" -n "$PRUNING_NS"
+    done
+  fi
+}
+
+rerun_job() {
+  for jobs in "${FailedJobs[@]}"; do
```
bmeng, line 36: If there are multiple failed jobs with the same type, like builds-pruner, I do not think we need to rerun each of them.
We need retry it once.

scripts/SREP/retry-pruning-cronjob/script.sh
Dee-6777: The Deployements_Pruner one is not being used.  Just declared it like Builds_Pruner. Shall I remove it? Or echo it?

scripts/SREP/retry-pruning-cronjob/script.sh
Dee-6777: Sure!

scripts/SREP/retry-pruning-cronjob/metadata.yaml
```diff
@@ -0,0 +1,26 @@
+file: script.sh
+name: fix-pruning-cron-job-error
```
bmeng, line 2:
```suggestion
name: retry-failed-pruning-job
```

scripts/SREP/retry-pruning-cronjob/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+description: Restart the failed jobs in openshift-sre-pruning namespace
```
bmeng, line 3:
```suggestion
description: Retry the failed jobs in openshift-sre-pruning namespace
```

PR #153, opened 2023-11-21
Fix Cronjob Command

Dee-6777:
Fix `oc get job` command to get jobs from a particular cronjob.


feichashao:
For a record, this PR fixes,
```
$ cat jobs.json|jq -r '.items[] | select(.metadata.ownerReferences[].name=="builds-pruner") | .metadata.name'
jq: error (at <stdin>:2006): Cannot iterate over null (null)
```
when there's no `ownerReferences`.

PR #157, opened 2023-12-04
OSD-19605 a script to extract email from pull secret

feichashao:
### What type of PR is this?

Feature

In some cases, SRE need to validate if the email in the pull secret matches the record in OCM.
Instead of pulling down the how secret file, we can use this script to only get the info needed for validation.

Part of [OSD-19605](https://issues.redhat.com//browse/OSD-19605)

The test result looks like this:
```
ocm backplane testjob logs openshift-job-dev-wnrbj 
siwu@redhat.com
```
It can be run on both classic OSD and Hypershift clusters.

### DIFF
--- a/scripts/SREP/get-pull-secret-email/script.sh
+++ b/scripts/SREP/get-pull-secret-email/script.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -o nounset
+set -o pipefail
+
+oc get secret pull-secret -n openshift-config -o json | jq -r '."data".".dockerconfigjson"' | base64 -d | jq -r '."auths"."cloud.openshift.com"."email"'
+
+exit 0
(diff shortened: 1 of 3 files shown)


scripts/SREP/get-pull-secret-email/script.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
bmeng, line 1: The existing osdctl subcommand https://github.com/openshift/osdctl/blob/master/cmd/cluster/validatepullsecret.go#L56 will check the ocm registry credential as well. It seems we cannot do the same in managed scripts?

feichashao: For the OCM part we still need to do that in `osdctl`.

bmeng:
Could you add a simple readme file for this script?

feichashao:
Added a README

PR #159, opened 2023-12-08
OSD-19604: Script to automate infranode resize without elevation

Dee-6777:
### What type of PR is this?

_feature_

It will help automate infra-node resize and without any elevation. SRE just needs to pass the Internal Cluster ID, hive name and the desired instance size.

_Resolves_ #[OSD-19604](https://issues.redhat.com/browse/OSD-19604)

This script runs in hive, so to test it in a staging cluster you need to create a hive namespace, install the hive operator, and then add the machinepools.

- [X] Validated the changes in a cluster
- [X] Included documentation changes with PR

### DIFF
--- a/scripts/SREP/infra-node-resize/metadata.yaml
+++ b/scripts/SREP/infra-node-resize/metadata.yaml
@@ -1,0 +1,28 @@
+file: script.sh
+name: infra-node-resize
+description: Resize infra node without elevation.
+author: dee-6777
+allowedGroups: 
+  - SREP
+rbac:
+    clusterRoleRules:
+        - verbs:
+            - "get"
+            - "list"
+            - "watch"
+            - "create"
+            - "delete"
+            - "patch"
+          apiGroups:
+            - "hive.openshift.io"
+          resources:
+            - "machinepools"
+            - "machinepools/scale"
+envs:
+  - key: "CLUSTER_ID"
+    description: "requires internal cluster ID"
+    optional: false
+  - key: "INSTANCE_SIZE"
+    description: "desired size"
+    optional: false
+language: bash
--- a/scripts/SREP/infra-node-resize/script.sh
+++ b/scripts/SREP/infra-node-resize/script.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+# This script checks if any tech preview features are enabled through feature sets.
+
+set -e
+set -o nounset
+set -o pipefail
+
+# fetch the infra machinepools
+get_machinepool() {
+    echo 'FETCHING INFRA MACHINEPOOLS...'
+    NAMESPACE=$(oc get ns | grep "$CLUSTER_ID" | awk '{print $1}')
+    oc -n "$NAMESPACE" get machinepool
+    INFRA_MACHINEPOOL=$(oc -n "$NAMESPACE"  get machinepool -o json | jq -r '.items[] | select(.spec.name=="infra") | .metadata.name')
+    oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r ".spec.platform.type"
+}
+
+# update the yaml file
+update_cluster_infra_yaml() {
+    echo 'UPDATING THE YAML FILE FOR TEMPORARY MACHINEPOOL...'
+    oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o yaml > /tmp/cluster-infra.yaml
+    yq -i 'del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)' /tmp/cluster-infra.yaml
+    cat /tmp/cluster-infra.yaml 
+    CLOUD=$(oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r '.spec.platform | keys_unsorted[0]')
+    yq -i ".metadata.name |= . + 2 | .spec.name |= . + 2 | .spec.platform.$CLOUD.type |= env(INSTANCE_SIZE)" /tmp/cluster-infra.yaml 
+}
+
+# create the temporary machinepool
+create_temp_machinepool() {
+    echo 'CREATING TEMPORARY MACHINEPOOL...'
+    oc -n "$NAMESPACE" create -f /tmp/cluster-infra.yaml
+    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
+    do
+        echo 'MACHINES ARE STILL NOT CORRECTLY PROVISIONED...'
+        timeout 10
+    done
+    oc -n "$NAMESPACE" get machinepool
+}
+
+# delete the original machinepool
+delete_original_machinepool() {
+    echo 'DELETING THE ORIGINAL MACHINEPOOL...'
+    oc scale --replicas 0 machinepool "$INFRA_MACHINEPOOL" -n "$NAMESPACE"
+    oc -n "$NAMESPACE" get machinepool
+    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
+    do
+        echo 'MACHINES ARE STILL NOT PROPERL DELETED...'
+        timeout 10
+    done
+    oc -n "$NAMESPACE" delete machinepool "$INFRA_MACHINEPOOL"
+    echo 'ORIGINAL MACHINEPOOLS WHICH ARE NOT RESIZED ARE DELETED...'
+    oc -n "$NAMESPACE" get machinepool
+}
+
+# update yaml file cluster 
+update_new_infra_yaml() {
+    echo 'UPDATING YAML FILE...'
+    yq eval '.metadata.name |= sub("2$"; "") | .spec.name |= sub("2$"; "")' -i /tmp/cluster-infra.yaml 
+    cat /tmp/cluster-infra.yaml
+}
+
+# create new machinepool
+create_new_machinepool() {
+    echo 'CREATING NEW MACHINEPOOL...'
+    oc -n "$NAMESPACE" create -f /tmp/cluster-infra.yaml
+    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
+    do
+        echo 'MACHINES ARE STILL NOT CORRECTLY PROVISIONED...'
+        timeout 10
+    done
+}
+
+# delete temporary machinepool 
+delete_temp_machinepool() {
+    echo 'DELETING TEMPORARY MACHINEPOOL...'
+    oc scale --replicas 0 machinepool "${INFRA_MACHINEPOOL}"2 -n "$NAMESPACE"
+    oc -n "$NAMESPACE" get machinepool
+    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
+    do
+        echo 'MACHINES ARE STILL NOT PROPERLY DELETED...'
+        timeout 10
+    done
+    oc -n "$NAMESPACE" delete machinepool "${INFRA_MACHINEPOOL}"2 
+    echo 'TEMPORARY MACHINEPOOLS ARE DELETED...'
+}
+
+# check the state of machinepools
+verify_machinepools() {
+    echo 'VERIFYING MACHINEPOOLS...'
+    echo 'FETCHING MACHINEPOOL NAME..'
+    INFRA_MACHINEPOOL_NAME=$(oc -n "$NAMESPACE"  get machinepool -o json | jq -r '.items[] | select(.spec.name=="infra") | .metadata.name')
+    echo "THIS INFRA MACHINEPOOL HAS BEEN RESIZED $INFRA_MACHINEPOOL_NAME"
+    INFRA_MACHINEPOOL_SIZE=$(oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o=jsonpath="{.spec.platform.$CLOUD.type}")
+    echo 'CHECKING IF IT HAS BEEN RESIZED CORRECTLY..'
+    if [[ "$INFRA_MACHINEPOOL_SIZE" == "$INSTANCE_SIZE" ]]; then
+        echo 'INFRA RESIZE HAS BEEN SUCCESSFULLY DONE..'
+    else
+        echo 'INFRA RESIZE UNSUCCESSFUL..'
+    fi
+}
+
+main() {
+  get_machinepool
+  update_cluster_infra_yaml
+  create_temp_machinepool
+  delete_original_machinepool
+  update_new_infra_yaml
+  create_new_machinepool
+  delete_temp_machinepool
+  verify_machinepools
+  echo "Succeed."
+  exit 0
+}
+
+main "$@"
--- a/scripts/SREP/infra-node-resize/README.md
+++ b/scripts/SREP/infra-node-resize/README.md
@@ -1,0 +1,16 @@
+# INFRA NODE RESIZE
+
+## Purpose
+
+This script will help automate infra node resize without having the need to elevate access permissions.
+
+## Usage
+To use this script you will need to first login to a Hive shard.
+This script runs on a production Hive shard and helps resize the infra nodes of the cluster to the desired size. Meaning, SREs do not have to elevate permissions to perform this operation.
+
+## Create the ManagedJob to resize infra-nodes
+```
+ocm backplane login <hive_shard>
+
+ocm backplane managedjob create SREP/infra-node-resize -p CLUSTER_ID=<Internal CLUSTER_ID> -p INSTANCE_SIZE=<desired instance size>
+```


scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# Infra-node-resize
+
+set -x
+set -e
+set -o nounset
+set -o pipefail
+
+# find hive_shard ID
```
MitaliBhalla, line 9:
```suggestion
# find hive_shard
```

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,53 @@
...
+CLUSTER_ID=ocm list clusters --parameter search="external_id is '${1}'" --columns id --padding 20 --no-headers
+HIVE_ID=$(ocm get /api/clusters_mgmt/v1/clusters/${CLUSTER_ID}/provision_shard | jq -r '.hive_config.server' | cut -d '.' -f 2)
```
MitaliBhalla, line 11: This can be changed to a better name as we'll have the name of the hive cluster and no such thing as 'HIVE ID'

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# update the yaml file
+update_cluster-infra-yaml() {
+yq '.items[] |= (del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)'
+yq `.metadata.name = .metadata.name | del(.metadata.name.)`
+}
+
+# silent the alerts
+silent_alerts() {
+    echo "To run this you need to have alerts script in your system"
```
MitaliBhalla, line 38:
```suggestion
    echo "To run this you need to have alerts script in your system"
```
I am confused as to what part of the code handles if the script is not present in someone's system?

MitaliBhalla (changes requested):

scripts/SREP/infra-node-resize/script.sh
Dee-6777: Added a condition to check that...will add further to download the file if not present.

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,109 @@
...
+    yq '.items[] |= (del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)'
+    yq `.metadata.name2 = .metadata.name | del(.metadata.name)`
+    yq `.spec.name2 = .spec.name | del(.spec.name)`
+}
+
+# silent the alerts
+silent_alerts() {
+    echo "To run this you need to have alerts script in your system"
+    echo "Checking alerts script..."
+    ocm backplane login $CLUSTER_ID
```
samanthajayasinghe, line 42: you may not able to execute `ocm backplane login` in managed script

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# delete temporary machinepool 
+delete_temp_machinepool() {
+    oc -n uhc-production-$CLUSTERID delete machinepool ${INFRA_MACHINEPOOL}2 
+}
+
+# send servicelog informing the resize
+send_servicelog() {
+  read -r JIRA_ID INSTANCE_TYPE JUSTIFICATION
+  osdctl servicelog post \
```
samanthajayasinghe, line 81: may not able to execute `osdctl` commands too

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# find hive shard 
+CLUSTER_UUID=$(oc get clusterversion version -o json | jq -r '.spec.clusterID')
+CLUSTER_ID=$(ocm list clusters --parameter search="external_id is '${CLUSTER_UUID}'" --columns id --padding 20 --no-headers)
+HIVE_NAME=$(ocm get /api/clusters_mgmt/v1/clusters/${CLUSTER_ID}/provision_shard | jq -r '.hive_config.server' | cut -d '.' -f 2)
+DESIRED_SIZE=x
+
+# get current infra-node size
+get_current_infra_size() {
+    ocm backplane login $HIVE_NAME
```
samanthajayasinghe, line 17: you are in the either hive or cluster , no need to login again

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,90 @@
...
+    oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r ".spec.platform"
+}
+
+# update the yaml file
+update_cluster_infra_yaml() {
+    oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o yaml > /tmp/cluster-infra.yaml
+    yq -i 'del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)' /tmp/cluster-infra.yaml
+    cat /tmp/cluster-infra.yaml 
+    CLOUD=$(oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r '.spec.platform | keys_unsorted[0]')
+    yq -i '.metadata.name |= . + "2" | .spec.name |= . + "2" | .spec.platform.'$CLOUD'.type |= env(INSTANCE_SIZE)' /tmp/cluster-infra.yaml
```
MitaliBhalla, line 23:
```suggestion
    yq -i '.metadata.name |= . + "2" | .spec.name |= . + "2" | .spec.platform."$CLOUD".type |= env(INSTANCE_SIZE)' /tmp/cluster-infra.yaml 
```

MitaliBhalla (changes requested):

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,90 @@
...
+    oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r ".spec.platform"
+}
+
+# update the yaml file
+update_cluster_infra_yaml() {
+    oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o yaml > /tmp/cluster-infra.yaml
+    yq -i 'del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)' /tmp/cluster-infra.yaml
+    cat /tmp/cluster-infra.yaml 
+    CLOUD=$(oc -n "$namespace" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r '.spec.platform | keys_unsorted[0]')
+    yq -i '.metadata.name |= . + "2" | .spec.name |= . + "2" | .spec.platform."$CLOUD".type |= env(INSTANCE_SIZE)' /tmp/cluster-infra.yaml
```
MitaliBhalla, line 23:
```suggestion
    yq -i ".metadata.name |= . + "2" | .spec.name |= . + "2" | .spec.platform."$CLOUD".type |= env(INSTANCE_SIZE)" /tmp/cluster-infra.yaml 
```

MitaliBhalla (changes requested):

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,90 @@
...
+
+main() {
+  get_current_infra_size
+  update_cluster_infra_yaml
+  create_temp_machinepool
+  delete_original_machinepool
+  update_new_cluster_infra_yaml
+  create_new_machinepool
+  delete_temp_machinepool
+  echo "Succeed."
```
devppratik, line 86: An echo statement on the entry of each step will be helpful to let the user know which step the script currently is. Also, would be helpful in case of script failure

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,90 @@
...
+# Infra-node-resize
+
+set -x
+set -e
+set -o nounset
+set -o pipefail
+
+# get current infra-node size
+get_current_infra_size() {
+    namespace=$(oc get ns | awk '{print $1}'| grep "$CLUSTER_ID")
```
devppratik, line 11: ```namespace=$(oc get ns | grep "$CLUSTER_ID" | awk '{print $1}' )```

To make sure we select the correct NS first then extract the value

devppratik:
I have tested the PR as follows

1. Create a stage cluster and install hive & related deployments on it
2. Create a NS similar to `uhc-staging-cluster-id`
3. Create the machine pool using a YAML
4. Run the script (Correctly executed)
6. The machine pool YAML is updated and size change reflects

However, requires a test on a staging hive cluster for LGTM

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+    echo 'UPDATING YAML FILE...'
+    yq eval '.metadata.name |= sub("2$"; "") | .spec.name |= sub("2$"; "")' -i /tmp/cluster-infra.yaml 
+    cat /tmp/cluster-infra.yaml
+}
+
+# create new machinepool
+create_new_machinepool() {
+    echo 'CREATING NEW MACHINEPOOL...'
+    oc -n "$NAMESPACE" create -f /tmp/cluster-infra.yaml
+    while [ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ];
```
xiaoyu74, line 65: I can see you use a `while loop with sleep` for waiting on machine provisioning and deletion. This could be a bit inefficient, since the loop keeps running as long as the condition `oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')` is true.  

In cases where machine provisioning or deletion takes longer than expected, or if there's an issue that prevents the condition from being achieved, the loop will continue running. This can lead to unexpectedly long wait times without any feedback to the user about what's happening ?  

So, may be better to introduce a timeout mechanism in **a separate function** to limit the number of checks to prevent the possible long waiting, WDYT ? cc: @bmeng

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+        sleep 10
+    done
+}
+
+# delete temporary machinepool 
+delete_temp_machinepool() {
+    echo 'DELETING TEMPORARY MACHINEPOOL...'
+    oc scale --replicas 0 machinepool "${INFRA_MACHINEPOOL}"2 -n "$NAMESPACE"
+    oc -n "$NAMESPACE" get machinepool
+    while [ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ];
```
xiaoyu74, line 77: Same as above

xiaoyu74:
Other than that, great work Deepti. :+1:

scripts/SREP/infra-node-resize/script.sh
Dee-6777: > So, may be better to introduce a timeout mechanism in a separate function to limit the number of checks to prevent the possible long waiting, also, adding error handling with an error message when timeout WDYT ? cc: @bmeng

Thanks, Shawn for the review! I can add a check function which will catch if there's any error but usually to get into ready state I had observed that it takes 10 mins. I can set the time-limit according to that. Thanks Again :)

scripts/SREP/infra-node-resize/script.sh
Dee-6777: Will fix this one too!

scripts/SREP/infra-node-resize/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+            - "delete"
+            - "patch"
+          apiGroups:
+            - "hive.openshift.io"
+          resources:
+            - "machinepools"
+            - "machinepools/scale"
+envs:
+  - key: "CLUSTER_ID"
+    description: "requires internal cluster id"
```
supreeth7, line 23:
```suggestion
    description: "requires internal cluster ID"
```

scripts/SREP/infra-node-resize/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+    optional: false
+  - key: "HIVE"
+    description: "requires HIVE name"
+    optional: false
+  - key: "INSTANCE_SIZE"
+    description: "requires size"
```
supreeth7, line 29:
```suggestion
    description: "required size"
```

supreeth7: Or better, use `desired size`

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+# Infra-node-resize
```
supreeth7, line 2: Let's make the header comments a bit more detailed?
Reference: https://github.com/openshift/managed-scripts/blob/main/scripts/CEE/check-tech-preview-features/script.sh#L2

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+    cat /tmp/cluster-infra.yaml 
+    CLOUD=$(oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r '.spec.platform | keys_unsorted[0]')
+    yq -i ".metadata.name |= . + 2 | .spec.name |= . + 2 | .spec.platform.$CLOUD.type |= env(INSTANCE_SIZE)" /tmp/cluster-infra.yaml 
+}
+
+# create the temporary machinepool
+create_temp_machinepool() {
+    echo 'CREATING TEMPORARY MACHINEPOOL...'
+    oc -n "$NAMESPACE" create -f /tmp/cluster-infra.yaml
+    while [ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ];
```
supreeth7, line 31: nit: `[` spawns a subshell, whereas `[[` does not ;)

```suggestion
    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
```

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+    do
+        echo 'Machines are still not correctly provisioned'
```
supreeth7, line 33: Why not make this all caps like the rest of the logs?

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+    done
+    oc -n "$NAMESPACE" get machinepool
+}
+
+# delete the original machinepool
+delete_original_machinepool() {
+    echo 'DELETING THE ORIGINAL MACHINEPOOL...'
+    oc scale --replicas 0 machinepool "$INFRA_MACHINEPOOL" -n "$NAMESPACE"
+    oc -n "$NAMESPACE" get machinepool
+    while [ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ];
```
supreeth7, line 44: ditto

```suggestion
    while [[ ! -z "$(oc get machinepool -n "$NAMESPACE" "$INFRA_MACHINEPOOL" -o=jsonpath='{.status.machineSets[0].errorMessage}')" ]];
```

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,98 @@
...
+}
+
+main() {
+  get_machinepool
+  update_cluster_infra_yaml
+  create_temp_machinepool
+  delete_original_machinepool
+  update_new_infra_yaml
+  create_new_machinepool
+  delete_temp_machinepool
```
supreeth7, line 93: Let's have another function; something like `verify_machinepools` that will fetch (get) the existing, ready, and provisioned machinepools and verify it's current instance size matches the desired size and prints the same to the console: 

`MACHINE | INSTANCE SIZE | DESIRED INSTANCE SIZE`

More of a sanity check.

supreeth7 (changes requested):
This is neat. I can already see this giving a good push to the ops routine.
Best to incorporate the suggested changes to have a more polished outcome.
Overall great work! :rocket:

scripts/SREP/infra-node-resize/script.sh
supreeth7: `timeout` can come in handy.

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,114 @@
...
+    INFRA_MACHINEPOOL=$(oc -n "$NAMESPACE"  get machinepool -o json | jq -r '.items[] | select(.spec.name=="infra") | .metadata.name')
+    oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o json | jq -r ".spec.platform.type"
+}
+
+# update the yaml file
+update_cluster_infra_yaml() {
+    echo 'UPDATING THE YAML FILE FOR TEMPORARY MACHINEPOOL...'
+    oc -n "$NAMESPACE" get machinepool "$INFRA_MACHINEPOOL" -o yaml > /tmp/cluster-infra.yaml
+    yq -i 'del(.metadata.creationTimestamp) | del(.metadata.finalizer) | del(.metadata.resourceVersion) | del(.metadata.generation) | del(.metadata.selfLink) | del(.metadata.uid) | del(.status)' /tmp/cluster-infra.yaml
+    cat /tmp/cluster-infra.yaml
```
bmeng, line 22: why need this step?

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,114 @@
...
+    oc -n "$NAMESPACE" delete machinepool "$INFRA_MACHINEPOOL"
+    echo 'ORIGINAL MACHINEPOOLS WHICH ARE NOT RESIZED ARE DELETED...'
+    oc -n "$NAMESPACE" get machinepool
+}
+
+# update yaml file cluster 
+update_new_infra_yaml() {
+    echo 'UPDATING YAML FILE...'
+    yq eval '.metadata.name |= sub("2$"; "") | .spec.name |= sub("2$"; "")' -i /tmp/cluster-infra.yaml 
+    cat /tmp/cluster-infra.yaml
```
bmeng, line 58: same, why do we cat the file here?
If we need to confirm the value, I think we need to add step like press `y` to continue?

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,114 @@
...
+    done
+    oc -n "$NAMESPACE" delete machinepool "$INFRA_MACHINEPOOL"
+    echo 'ORIGINAL MACHINEPOOLS WHICH ARE NOT RESIZED ARE DELETED...'
+    oc -n "$NAMESPACE" get machinepool
+}
+
+# update yaml file cluster 
+update_new_infra_yaml() {
+    echo 'UPDATING YAML FILE...'
+    yq eval '.metadata.name |= sub("2$"; "") | .spec.name |= sub("2$"; "")' -i /tmp/cluster-infra.yaml
```
bmeng, line 57: will it be better if we build the yaml for the new machinepool separately?
so we do not need to update the same file twice.

bmeng: means we can build two yaml from the func `update_cluster_infra_yaml` maybe change the name to `build_cluster_infra_yaml`
one for the name with prefix -2 and one for the original name, both have the same size.
And in the following steps, we just need to create `-2` and delete original and create new-original and delete `-2`.

scripts/SREP/infra-node-resize/script.sh
Dee-6777: We can't interact using managedscript :disappointed:

scripts/SREP/infra-node-resize/script.sh
Dee-6777: To show if it has been correctly updated

scripts/SREP/infra-node-resize/script.sh
Dee-6777: Good idea! But I think we can keep it as it is, both are in separate functions and changing the yaml files separately. It won't change the code complexity.

scripts/SREP/infra-node-resize/script.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+# This script checks if any tech preview features are enabled through feature sets.
```
feichashao, line 2: This is not relevant?

Dee-6777: Changing

feichashao:
Tested in a mock hive environment and it works good.

PR #166, opened 2023-12-27
Feat/retrieve pdbs

danifernandezs:
### What type of PR is this?

Feature

### What does this PR do / Why do we need it?

This PR introduces a script to review the deployed PodDisruptionBudgets (PDBs) across cluster namespaces.
It includes a "restrictive" option to retrieve only the restrictive PDBs in all namespaces.

N/A

This script is a part of the Cluster Health Check steps.

- [X] Validated the changes in a cluster
- [X] Included documentation and examples

### DIFF
--- a/scripts/CEE/retrieve-pdbs/script.sh
+++ b/scripts/CEE/retrieve-pdbs/script.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+set -e
+set -o nounset
+set -o pipefail
+
+# Function to retrieve all namespaces starting with openshift
+get_openshift_namespaces() {
+    oc get namespaces -o json | jq -r '.items[].metadata.name' | grep '^openshift'
+}
+
+# Function to retrieve and display PDBs for a given namespace
+get_and_display_pdbs() {
+    local namespace="$1"
+    echo "---------------------------------"
+    echo "PDBs in namespace: $namespace"
+    echo ""
+    oc get poddisruptionbudget -n "$namespace"
+}
+
+# Check if the 'namespace' variable is provided
+if [ -z "${namespace:-}" ]; then
+    echo "No 'namespace' provided. Retrieving PDBs from all openshift namespaces."
+    openshift_namespaces=( $(get_openshift_namespaces) )
+
+    # Loop through each openshift namespace and display PDBs
+    for namespace in "${openshift_namespaces[@]}"; do
+        echo ""
+        get_and_display_pdbs "$namespace"
+    done
+
+elif [ "${namespace}" = "--all" ]; then
+    echo "Retrieving PDBs from all namespaces in the cluster."
+    for namespace in $(oc get namespaces -o jsonpath='{.items[*].metadata.name}'); do
+        echo ""
+        get_and_display_pdbs "$namespace"
+    done
+
+elif [ "${namespace}" = "--restrictive" ]; then
+    echo "Checking for restrictive PDBs in all namespaces in the cluster."
+
+    restrictive_found=false
+
+    for namespace in $(oc get namespaces -o jsonpath='{.items[*].metadata.name}'); do
+        echo ""
+        restrictive_pdbs=$(oc get poddisruptionbudget -n "$namespace" -o json | jq -r '.items[] | select(.spec.maxUnavailable == 0 or .spec.maxUnavailable == "0%") | "\(.metadata.name)\t-->\tmaxUnavailable: \(.spec.maxUnavailable)"')
+
+        if [ -n "${restrictive_pdbs}" ]; then
+            echo "---------------------------------"
+            echo "Restrictive PDBs found in namespace: $namespace"
+            echo "${restrictive_pdbs}"
+            restrictive_found=true
+        else
+            echo "---------------------------------"
+            echo "No restrictive PDBs found in namespace: $namespace"
+        fi
+    done
+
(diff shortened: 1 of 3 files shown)


scripts/CEE/retrieve-pdbs/script.sh
```diff
@@ -0,0 +1,78 @@
...
+    oc get poddisruptionbudget -n "$namespace"
+}
+
+# Check if the 'namespace' variable is provided
+if [ -z "${namespace:-}" ]; then
+    echo "No 'namespace' provided. Retrieving PDBs from all openshift namespaces."
+    openshift_namespaces=( $(get_openshift_namespaces) )
+
+    # Loop through each openshift namespace and display PDBs
+    for namespace in "${openshift_namespaces[@]}"; do
```
bmeng, line 27: Use a for loop is kind of time consuming as well as api call consuming.
Do we have the special use case to split the namespaces like this?
Or can we just simplify the script to only retrieve the PDB for all namespaces?

hectorakemp:
@danifernandezs did you have any further updates regarding [this](https://github.com/openshift/managed-scripts/pull/166#discussion_r1514035779) comment?

danifernandezs:
Hi @hectorakemp & @bmeng 

The use cases of the loop that I thought of are.

- If we (MCS) don't declare any namespaces, we only check those from openshift-* (I need to loop with the retrieved namespaces), to avoid PDBs from customer namespaces.
- Show/Print PDBs in the same way as another script (the event one)

scripts/CEE/retrieve-pdbs/script.sh
```diff
@@ -0,0 +1,78 @@
...
+    oc get namespaces -o json | jq -r '.items[].metadata.name' | grep '^openshift'
+}
+
+# Function to retrieve and display PDBs for a given namespace
+get_and_display_pdbs() {
+    local namespace="$1"
+    echo "---------------------------------"
+    echo "PDBs in namespace: $namespace"
+    echo ""
+    oc get poddisruptionbudget -n "$namespace"
```
feichashao, line 18: I have some concerns as it gets data from customer namespaces.  I would like to understand more on the purpose on getting the PDB. 
If that's for health check, I think it is better to only output relevant data, eg, only output "It has bad PDBs in customer namespace"/"All PDBs look good in customer namespace".   Then guide customer to check the PDBs with certain commands.

feichashao:
> Hi @hectorakemp & @bmeng
> 
> The use cases of the loop that I thought of are.
> 
>     * If we (MCS) don't declare any namespaces, we only check those from openshift-* (I need to loop with the retrieved namespaces), to avoid PDBs from customer namespaces.
> 
>     * Show/Print PDBs in the same way as another script (the event one)

Thanks for the explanation! I think it is reasonable to loop through all the openshift-* namespaces. The number of openshift-* is not large so the api requests should be manageable. 

For the "--all", I think it makes more sense to do a `oc get pdb -A` so that it only needs one single API request. Also I left a comment above regarding getting the pdb data in customer namespaces. I would like to know your thoughts.

Thanks!

scripts/CEE/retrieve-pdbs/script.sh
danifernandezs: Hi @feichashao 

Let me share more context behind the idea of the initial script development.

- In proactive cases, the MCS team is interested in knowing if there are restrictive PDBs that could affect the nodes roll-out
- By default, we only read openshift-* to avoid reading data from the custiomer namespaces
- Added --all or specific namespaces option to debug certain cases where customers have pod rotation or workload migration issues, simply to verify these configurations without requesting a namespace inspection. (example, new machinePool that requires migrating the entire workload)


For MCS team, the --restrictive would be enough.
The rest are additions that should be able to be dispensed with or open an OHSS for troubleshooting.

danifernandezs:
I can modify the "all namespaces" option to use the -A flag.
The motivation behind individual API requests is to keep the output format consistent with the rest of the options.

danifernandezs:
Hi @xiaoyu74 

Why this PR was flagged to be on hold???

bmeng:
@jaybeeunix Could you help take a look of this script from compliance pov? Thanks.

bmeng:
/unhold
But it needs a review from security team

feichashao:
@sam-nguyen7 regarding the customer data concern, what about the approach of, 
> If that's for health check, I think it is better to only output relevant data, eg, only output "It has bad PDBs in customer namespace"/"All PDBs look good in customer namespace".

In this case, the script will read the PDB customer namespace, but it won't output sensitive data of customer namespace.

xiaoyu74:
@jaybeeunix @sam-nguyen7 - Any feedback regarding the security concerns?

sam-nguyen7:
> @jaybeeunix @sam-nguyen7 - Any feedback regarding the security concerns?

@xiaoyu74 Sorry for not getting back to you sooner. Looking at this closer and at more osdctl logic, I'm +1 to reading the PDB in customer's namespace AND the name of the PDB as I think this will give better support. For compliance, we can justify this as an approved and procedural part of our operations on a given cluster through our tooling.

Tessg22:
lifecycle/stale

danifernandezs:
I’m closing this pull request as no decision has been made after a year.
Additionally, I no longer work at Red Hat, so merging this code is no longer relevant to me.

Thank you for the opportunity to contribute.

PR #169, opened 2024-02-13
Kubelet restart

madhusudanupadhyay:
### What type of PR is this?

Managed Script addition for Kubelet restart

- Restart Kubelet on Master node. 
-  Customer with cluster admin perms can do it but since the action needs to be on Master node, a preferable way would be doing it via managed script by MCS secondment engineers
- Issues like: pod stuck in containerUnknown Status, bugs that require kubelet restarts to unblock upgrades etc. 

From one of the use cases, an example could be: https://access.redhat.com/solutions/6976343

Reducing an OHSS and turn around time for customers by letting MCS secondment engineers quickly restart the kubelet in scenarios where required.


scripts/CEE/kubelet-restart/script.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+set -e
```
clcollins, line 3:
```suggestion
set -o errexit
```

...if you want it to match :D

scripts/CEE/kubelet-restart/README.md
```diff
@@ -0,0 +1,13 @@
+# Restart Kubelet on Master Node
```
clcollins, line 1: How about generic "Restart Kubelet on a Node"

scripts/CEE/kubelet-restart/script.sh
madhusudanupadhyay: Done! Thanks :)

scripts/CEE/kubelet-restart/README.md
madhusudanupadhyay: Changes made for it to not be node-specific.! Thanks!

scripts/CEE/kubelet-restart/README.md
```diff
@@ -0,0 +1,13 @@
+# Restart Kubelet on Node
+
+This will script will restart kubelet on Node. Scenarios where a restart might be required:
```
bmeng, line 3:
```suggestion
This script will restart kubelet on Node. Scenarios where a restart might be required:
```

scripts/CEE/kubelet-restart/metadata.yaml
```diff
@@ -0,0 +1,48 @@
+file: script.sh
+name: restart-kubelet
+description: Restart kubelet for a specific node
+author: Madhusudan Upadhyay
+allowedGroups:
+  - SREP
+  - LPSRE
+  - CEE
+rbac:
+  roles:
```
bmeng, line 10: why do we need both the namespaced role and cluster role here?

bmeng:
@jaybeeunix would you mind to take a look at this script from compliance point of view?

jaybeeunix:
it looks good from my perspective.

PR #173, opened 2024-04-16
Managed Script for "Restart default dns pods in openshift-dns namespace"

diakovnec:
### What type of PR is this?

This script will trigger a rollout of daemonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  

We have a use case scenarios where a customer is required to restart dns default e.g. after applying a dns forwarding (if this script is considered as a valid then it can be applied to apiserver etc which will reduce the number of requests for SRE) pods and can't do it to multiple reasons (e. g. not having enough permissions, not being able to log in or not having enough expertise). Having this script will let MCS to perform this on behalf of the customer. This script will be run by the CRU member only.

### DIFF
--- a/scripts/CEE/restart-dns-default/README.md
+++ b/scripts/CEE/restart-dns-default/README.md
@@ -1,0 +1,14 @@
+# Restart default dns pods in openshift-dns namespace
+
+## Desctiption
+ 
+This cript will trigger a rollout of deamonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
+
+## Usage
+
+```bash
+ocm backplane managedjob create CEE/restart-dns-default -p NAMESPACE=openshift-dns
+```
+
+
+
--- a/scripts/CEE/restart-dns-default/README.md
+++ b/scripts/CEE/restart-dns-default/README.md
@@ -2,7 +2,7 @@
 
 ## Desctiption
  
-This script will trigger a rollout of deamonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
+This script will trigger a rollout of daemonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
 
 ## Usage
 
--- a/scripts/CEE/restart-dns-default/README.md
+++ b/scripts/CEE/restart-dns-default/README.md
@@ -7,7 +7,7 @@ This script will trigger a rollout of daemonset, controlling dns default pods. S
 ## Usage
 
 ```bash
-ocm backplane managedjob create CEE/restart-dns-default -p NAMESPACE=openshift-dns
+ocm backplane managedjob create CEE/restart-dns-default
 ```
 
 
--- a/scripts/CEE/restart-dns-default/README.md
+++ b/scripts/CEE/restart-dns-default/README.md
@@ -2,7 +2,7 @@
 
 ## Desctiption
  
-This cript will trigger a rollout of deamonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
+This script will trigger a rollout of deamonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
 
 ## Usage
 
--- a/scripts/CEE/restart-dns-default/script.sh
+++ b/scripts/CEE/restart-dns-default/script.sh
@@ -5,6 +5,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+NAMESPACE=openshift-dns
+
 CURRENTDATE=$(date +"%Y-%m-%d %T")
 
 ## validate input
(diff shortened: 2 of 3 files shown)


scripts/CEE/restart-dns-default/README.md
```diff
@@ -0,0 +1,14 @@
+# Restart default dns pods in openshift-dns namespace
+
+## Desctiption
+ 
+This script will trigger a rollout of daemonset, controlling dns default pods. Such action might be required when a customer is not able to perform it themself due to not having enough permissions etc, but dns-default pods need to be restarted.  
+
+## Usage
+
+```bash
+ocm backplane managedjob create CEE/restart-dns-default -p NAMESPACE=openshift-dns
```
xiaoyu74, line 10: I assume the script is solely intended for restarting dns pods within the `openshift-dns`, is that necessary to specify it as a parameter `-p NAMESPACE` or just hard-code in the script?

scripts/CEE/restart-dns-default/script.sh
```diff
@@ -0,0 +1,88 @@
...
+
+}
+
+verify_dns_pods_are_ready(){
+
+echo -e "\nVerifying that pods are in Running state..."
+
+sleep 30
+
+if [[ $(oc wait pods -n openshift-dns -l dns.operator.openshift.io/daemonset-dns=default --for=condition=Ready) ]]; then
```
xiaoyu74, line 68: Looks you are trying to use  `oc wait` to ensure that all DNS pods are in the "Ready" state. However, the handling around this command doesn't properly capture the outcome when the condition isn't met (The logic doesn't seem to handle `if the pods do not reach the "Ready" state`).

scripts/CEE/restart-dns-default/script.sh
```diff
@@ -0,0 +1,88 @@
...
+        
+        sleep 30	
+	echo -e "\nAll pods are in the Running state\n";
+	oc get pods -n openshift-dns -l dns.operator.openshift.io/daemonset-dns=default
+	exit 1
```
xiaoyu74, line 73: This `exit 1` looks redundant if verifying that all pods are running.  Normally, an exit stratus `exit 1` indicates an error, but in your case it doesn't handle the error condition as I mentioned in the above comment but still signal a failure `exit 1` status.

scripts/CEE/restart-dns-default/script.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "Job finished at $CURRENTDATE"
+}
+
+
+
+## Function for daemonsent rollout start
+restart_dns_pods(){
+
+    echo -e "\nRestarting dns_pods in \"${NAMESPACE}\" namespace. Wait until rollout reaches 5 out 5"
+
```
xiaoyu74, line 34: Looks you hard coded the `openshift-dns` but also use the variable `$NAMESPACE` which introduced the discrepancies. You can define the variable at the top of the script and assign the value `openshift-dns` to make the use consistent.

xiaoyu74:
I can see you submitted the PR from your local `main` branch which is not the best practice, better to avoid submitting from main but creating a feature or topic branch (e.g `restart_dns_pods`) for the changes and submit a PR from there.

xiaoyu74:
Also, please test your managed script against an actual OSD/ROSA cluster in staging environment to make sure it worked properly.

scripts/CEE/restart-dns-default/README.md
diakovnec: For this particular script we definitely can have it hard-coded. The idea was if this script is considered as a valid option for Managed Scripts and proven working then to introduce/or develop more sophisticated script  that will target different `openshift-* `namespaces e.g. `openshift-kube-apiserver` or `openshift-ovn-kubernetes` and thats where we pass a namespace as a parameter `-p NAMESPACE. ` I will follow your suggestion and hard code `var` in the script

scripts/CEE/restart-dns-default/script.sh
diakovnec: Changed the logic by following the example of existing script 

[delete-worker-node](https://github.com/openshift/managed-scripts/blob/accd40c26f25b7adc538c2b1dfbea2f9ea743513/scripts/CEE/delete-worker-node/script.sh#L56)

```
verify_dns_pods_are_ready(){

echo -e "\nVerifying that pods are in Running state..."

pod_status=$(oc get pod -l dns.operator.openshift.io/daemonset-dns=default -ojson | jq '[.items[]?.status.phase!="Running"] | any ')

    if [ "$pod_status" == "false" ]; then
        echo -e "\n[SUCCESS] the rest of the pods are in Running state\n"
        oc get pods -n openshift-dns -l dns.operator.openshift.io/daemonset-dns=default
    else
        sleep 5
        echo -e "\n[Error], there are pods are in not Running state, please, check them \n"
        oc get pods -n openshift-dns -l dns.operator.openshift.io/daemonset-dns=default
    fi
}
```

scripts/CEE/restart-dns-default/script.sh
diakovnec: removed redundant `exit 1`

scripts/CEE/restart-dns-default/script.sh
diakovnec: hard-coded and removed from metadata

diakovnec: @xiaoyu74 
regarding the comment 

`I can see you submitted the PR from your local main branch which is not the best practice, better to avoid submitting from main but creating a feature or topic branch (e.g restart_dns_pods) for the changes and submit a PR from there.`

Do you want me to close this PR and create a new one with the new branch, or leave as it is?

xiaoyu74: @diakovnec - Sorry for my late response. Yeah, please create a new one with all the updates in the script and submit from your local feature/topic branch. Thank you.

PR #176, opened 2024-04-24
migrate dp-012-increasing-observability-operator-pvc-size.md to managed scripts

boranx:
### What type of PR is this?

feature

Fix: https://issues.redhat.com/browse/MTSRE-1769

This needs to be executed using lPSRE group permissions that are assigned to associates

- [ ] Validated the changes in a cluster (not yet)

### DIFF
--- a/scripts/LPSRE/increasing-observability-operator-pvc-size/metadata.yaml
+++ b/scripts/LPSRE/increasing-observability-operator-pvc-size/metadata.yaml
@@ -1,0 +1,41 @@
+file: script.sh
+name: increasing-observability-operator-pvc-size # name should be the same as the subdir, eg: SRE/example in this case
+description: Increase the size of the PVCs used by the Observability Operator running in rhacs-observability namespace
+author: bseref
+allowedGroups: 
+  - LPSRE
+rbac:
+    roles: []
+    clusterRoleRules:
+    - verbs:
+      - "get"
+      - "list"
+      - "patch"
+      - "watch"
+      apiGroups:
+      - "apps"
+      resources:
+      - deployments
+      - statefulsets
+    - verbs:
+      - "get"
+      - "list"
+      apiGroups:
+      - ""
+      resources:
+      - pods
+      - pods/log
+    - verbs:
+      - "get"
+      - "list"
+      - "patch"
+      apiGroups:
+      - ""
+      resources:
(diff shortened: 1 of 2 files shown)


scripts/LPSRE/increasing-observability-operator-pvc-size/metadata.yaml
```diff
@@ -0,0 +1,41 @@
+file: script.sh
+name: increasing-observability-operator-pvc-size # name should be the same as the subdir, eg: SRE/example in this case
+description: Increase the size of the PVCs used by the Observability Operator running in rhacs-observability namespace
+author: bseref
+allowedGroups: 
+  - LPSRE
+rbac:
+    roles: []
+    clusterRoleRules:
```
bmeng, line 9: Should be able to use role/rolebinding here instead of cluster level, as seems all the operations are happened in `rhacs-observability` namespace

scripts/LPSRE/increasing-observability-operator-pvc-size/script.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+# This job outlines the procedure for increasing the PVC size of an RHACS (Red Hat Advanced Cluster Security) instance. 
+# It includes steps for inspecting logs, retrieving the current PVC, editing the PVC size, and verifying the resize process.
+set -euxo pipefail
+
+if [[ -z "${NEW_SIZE}" ]]; then
```
bmeng, line 6: small suggestion, maybe validate the NEW_SIZE before apply to avoid human error.
like, the NEW_SIZE should be larger than existing, or the given format is correct via regex.

scripts/LPSRE/increasing-observability-operator-pvc-size/metadata.yaml
boranx: nice suggestion! I've limited the permissions to namespace scope only

scripts/LPSRE/increasing-observability-operator-pvc-size/script.sh
boranx: makes sense, I've added both format check using regex and size check. Though calculating the sizing is bringing some overhead to the readability

Nanyte25 (approved):
/approved

PR #178, opened 2024-04-29
Delete service account in the default namespace 

diakovnec:
### What type of PR is this?

This script will remove a service account that was deployed in the default namespace as a part of custom workload. The service account needs to be passed as an argument SERVICE_ACCOUNT 

From time to time we have such requests and having this script will lower the load to SRE to look into such tasks/

https://issues.redhat.com/browse/OHSS-34289

All check have been done in staging and shellcheck used to check script itself


diakovnec:
Closing this PR as I've made a mistake and created branch not from main and it included unnecessary files

PR #179, opened 2024-04-29
Delete service account in the default namespace

diakovnec:
What type of PR is this?

This script will remove a service account that was deployed in the default namespace as a part of custom workload. The service account needs to be passed as an argument SERVICE_ACCOUNT

From time to time we have such requests and having this script will lower the load to SRE to look into such tasks.

What this PR does / Why we need it?

https://issues.redhat.com/browse/OHSS-34289

Pre-checks (if applicable)

All check have been done in staging and shellcheck used to check script itself

### DIFF
--- a/scripts/CEE/delete-service-account/script.sh
+++ b/scripts/CEE/delete-service-account/script.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+set -e
+set -o nounset
+set -o pipefail
+
+CURRENTDATE=$(date +"%Y-%m-%d %T")
+
+if [[ -z "${SERVICE_ACCOUNT}" ]]; then
+    echo "Variable SERVICE_ACCOUNT cannot be blank"
+    exit 1
+fi
+
+start_job(){
+    echo "Job started at $CURRENTDATE"
+    echo ".................................."
+    echo
+}
+
+finish_job(){
+    echo
+    echo ".................................."
+    echo "Job finished at $CURRENTDATE"
+}
+
+
+## Get info about existing service accounts
+get_existing_sa(){
+
+    echo -e "\nVerifying all existing service accounts in the default namespace\n"
+    oc get sa -n default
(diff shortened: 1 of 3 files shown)


scripts/CEE/delete-service-account/script.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+set -e
+set -o nounset
+set -o pipefail
+
+CURRENTDATE=$(date +"%Y-%m-%d %T")
```
xiaoyu74, line 7: nit: This variable `CURRENTDATE` you've set is to log when the job starts and finishes. However, it's initialized only  at the beginning of the script. If the script takes longer time to run, the finish time may be inaccurate.

So, probably better to move the date command directly into the `start_job` and `finish_job` functions like:
```
start_job(){
    local start_date=$(date +"%Y-%m-%d %T")
    ...
}

finish_job(){
    local finish_date=$(date +"%Y-%m-%d %T")
    ...
}
```

bmeng:
delete service account is not a regular operation IMO, I would not recommend to do it via managed scripts

xiaoyu74:
One thing I am not very sure if MCS/CEE should be granted the `deletion permission` for the SA in the `default` namespace? cc: @bmeng @feichashao

xiaoyu74:
> delete service account is not a regular operation IMO, I would not recommend to do it via managed scripts

left a comment almost the same time, got same concerns

diakovnec:
So I assume the answer is "REJECTED"?

feichashao:
Hey @diakovnec , as per offline discussion, let's consider this as rejected. We can find alternative ways when we need it.

diakovnec:
This was discussed https://redhat-internal.slack.com/archives/C016S65RNG5/p1715056594087299 and decision made to REJECT due to compliance policies

PR #182, opened 2024-05-07
Use specific UBI version, and update yq version

fahlmant:
### What type of PR is this?
Version bumps to UBI and yq

Pins UBI to a specific version instead of using latest, and updates the yq version to get the latest CVE patches. 

https://issues.redhat.com/browse/OSD-22358


rafael-azevedo (approved):
/lgtm

PR #187, opened 2024-05-28
Adding script files to collect and filter cluster autoscaler logs

kkulshreshth:
### What type of PR is this?
A feature addition for MCS team that will enable them to collect and filter verbose cluster autoscaler logs.

The script will be responsible for collection and filtering verbose logs from cluster autoscaler pod. We get a lot of cases where the customer reports issues with autoscaling nodes in the cluster (for ex: underutilized nodes are not scaling down). With current logs verbosity, it is not possible to find out the exact reason for why the nodes were not scaling down/up. On the other hand, setting permanent high value for logVerbosity aggressively uses the  storage. The script will update the verbosity, collect the verbose logs for 6 minutes and reset the logVerbosity value. The script will also filter the collected based on node names. 

Additionally, we will be using this script to completely automated troubleshooting for cluster autoscaling cases using our CLI troubleshooting tool inceptor.  

https://issues.redhat.com/browse/OSD-22869

### DIFF
--- a/scripts/CEE/collect-verbose-ca-logs/script.sh
+++ b/scripts/CEE/collect-verbose-ca-logs/script.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+# Set variables
+GREEN="\033[32m"
+RESET="\033[0m"
+NAMESPACE="openshift-machine-api"
+CA_NAME="default"
+
+# Get current logVerbosity value
+current_log_verbosity=$(oc get ca $CA_NAME -o jsonpath='{.spec.logVerbosity}')
+
+# Check if current_log_verbosity is empty
+if [ -z "$current_log_verbosity" ]; then
+  echo "Failed to retrieve current logVerbosity value."
+  exit 1
+fi
+
+echo
+echo "CURRENT LOG VERBOSITY = $current_log_verbosity"
+echo
+
+# Update logVerbosity to 6
+oc patch ca $CA_NAME --type='json' -p='[{"op": "replace", "path": "/spec/logVerbosity", "value": 6}]'
+
+updated_log_verbosity=$(oc get ca $CA_NAME -o jsonpath='{.spec.logVerbosity}')
+echo
+echo "UPDATED LOG VERBOSITY = $updated_log_verbosity"
+echo
+
+# Wait for the update to take effect
+echo "Waiting for the log verbosity to update..."
+sleep 30
+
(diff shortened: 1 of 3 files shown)


scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+oc patch ca $CA_NAME --type='json' -p='[{"op": "replace", "path": "/spec/logVerbosity", "value": 6}]'
+
+updated_log_verbosity=$(oc get ca $CA_NAME -o jsonpath='{.spec.logVerbosity}')
+echo
+echo "UPDATED LOG VERBOSITY = $updated_log_verbosity"
+echo
+
+# Wait for the update to take effect
+echo "Waiting for the log verbosity to update..."
+sleep 30
```
devppratik, line 32: Instead of using sleep, we should ideally use read the current value from the resource YAML  to make sure it was updated.
The logic here would be check the value every 5-10s until the value gets updated

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# Revert logVerbosity to previous value
+oc patch ca $CA_NAME --type='json' -p="[{'op': 'replace', 'path': '/spec/logVerbosity', 'value': $current_log_verbosity}]"
+
+# Wait for the update to take effect
+echo
+echo "Waiting for the log verbosity to be reverted back..."
+echo
+
+sleep 30
```
devppratik, line 76: Similar case here - Verifying the actual value rather than using sleep

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+# Get the name of the Cluster Autoscaler pod
+CA_POD=$(oc get pods -n openshift-machine-api | grep cluster-autoscaler-default | awk '{print $1}')
+
+# Collect logs for the next 6 minutes
+echo
+echo "Collecting logs for the next 6 minutes..."
+echo
+
+sleep 360
+
```
devppratik, line 43: I am thinking the duration for 6 minutes is quite large without any update to the user. It would better if there is a constant Info printed out. Example
```
Waiting 6 minutes
Waiting 5m 30s...
Waiting 5minutes for log collection
```

PR #188, opened 2024-05-28
Feature cluster autoscaler logs

kkulshreshth:
### What type of PR is this?

A feature addition for MCS team that will enable them to collect and filter verbose cluster autoscaler logs.

The script will be responsible for collection and filtering verbose logs from cluster autoscaler pod. We get a lot of cases where the customer reports issues with autoscaling nodes in the cluster (for ex: underutilized nodes are not scaling down). With current logs verbosity, it is not possible to find out the exact reason for why the nodes were not scaling down/up. On the other hand, setting permanent high value for logVerbosity aggressively uses the storage. The script will update the verbosity, collect the verbose logs for 6 minutes and reset the logVerbosity value. The script will also filter the collected based on node names.

Additionally, we will be using this script to completely automated troubleshooting for cluster autoscaling cases using our CLI troubleshooting tool inceptor.

https://issues.redhat.com/browse/OSD-22869

### DIFF
--- a/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
+++ b/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
@@ -33,3 +33,4 @@ rbac:
         resources:
         - "nodes"
 language: bash
+customerDataAccess: false
--- a/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
+++ b/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
@@ -29,7 +29,7 @@ rbac:
         - "get"
         - "list"
         apiGroups:
-        - "metrics.k8s.io"
+        - ""
         resources:
         - "nodes"
       - verbs:
--- a/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
+++ b/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
@@ -1,0 +1,42 @@
+file: script.sh
+name: collect-verbose-ca-logs
+description: Script responsible for collecting verbose logs from cluster-autoscaler
+author: Kushagra Kulshreshtha
+allowedGroups:
+  - CEE
+  - SREP
+rbac:
+  roles:
+    - namespace: "openshift-machine-api"
+      rules:
+        - verbs:
+          - "get"
+          - "logs"
+          apiGroups:
+          - "clusterautoscaler.autoscaling.openshift.io"
+          resources:
+          - "pods"
+          - "pods/attach"
+  clusterRoleRules:
+      - verbs:
+        - "get"
+        - "patch"
+        apiGroups:
+        - "autoscaling.openshift.io"
+        resources:
+        - "clusterautoscalers"
+      - verbs:
+        - "get"
+        - "list"
+        apiGroups:
+        - "metrics.k8s.io"
+        resources:
+        - "nodes"
+      - verbs:
+        - "get"
+        - "list"
+        apiGroups:
+        - ''
+        resources:
+        - "nodes"
+language: bash
--- a/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
+++ b/scripts/CEE/collect-verbose-ca-logs/metadata.yaml
@@ -32,11 +32,4 @@ rbac:
         - ""
         resources:
         - "nodes"
-      - verbs:
-        - "get"
-        - "list"
-        apiGroups:
-        - ''
-        resources:
-        - "nodes"
 language: bash
--- a/scripts/CEE/collect-verbose-ca-logs/script.sh
+++ b/scripts/CEE/collect-verbose-ca-logs/script.sh
@@ -29,7 +29,27 @@ echo
 
 # Wait for the update to take effect
 echo "Waiting for the log verbosity to update..."
-sleep 30
+
+# Sleep for 5 seconds so that the new pod name gets reflected
+sleep 5
+
+CA_POD=$(oc get pods -n openshift-machine-api | grep cluster-autoscaler-default | awk '{print $1}')
+echo "POD Name: $CA_POD"
+
+while true; do
+  # Get the name and status of the cluster-autoscaler pod
+  POD_STATUS=$(oc get pod "$CA_POD" -n $NAMESPACE -o jsonpath='{.status.phase}')
+
+  # Check if the pod status is "Running"
+  if [ "$POD_STATUS" == "Running" ]; then
+    echo "The cluster-autoscaler pod is now 'Running'."
+    break
+  fi
+
+  # Print the current status and wait before rechecking
+  echo "Current status: $POD_STATUS. Waiting..."
+  sleep 5
+done
 
 # Get the name of the Cluster Autoscaler pod
 CA_POD=$(oc get pods -n openshift-machine-api | grep cluster-autoscaler-default | awk '{print $1}')
@@ -39,6 +59,7 @@ echo
 echo "Collecting logs for the next 6 minutes..."
 echo
 
+# Sleep for 6 minutes to let CA pod generate logs
 sleep 360
 
 echo "---------------------"
@@ -47,7 +68,7 @@ echo "---------------------"
 echo
 CA_LOGS=$(oc logs -n $NAMESPACE "$CA_POD" --since=6m)
 
-#Collect the list of nodes from the cluster
+# Collect the list of nodes from the cluster
 node_names=$(oc get nodes -o jsonpath='{.items[*].metadata.name}')
 
 for node in $node_names; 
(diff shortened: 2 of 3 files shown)


scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+oc patch ca $CA_NAME --type='json' -p='[{"op": "replace", "path": "/spec/logVerbosity", "value": 6}]'
+
+updated_log_verbosity=$(oc get ca $CA_NAME -o jsonpath='{.spec.logVerbosity}')
+echo
+echo "UPDATED LOG VERBOSITY = $updated_log_verbosity"
+echo
+
+# Wait for the update to take effect
+echo "Waiting for the log verbosity to update..."
+sleep 30
```
feichashao, line 32: In https://github.com/openshift/managed-scripts/pull/187  @devppratik mentioned the idea about instead of blindly sleeping, we might wait for some certain conditions.  Any thought about that?

I haven't tried changing the verbosity by myself, could you help me to understand what it is waiting for? I assume it is waiting for the auto scaler pods to redeploy?

feichashao:
@kkulshreshth Thank you for the PR!  I added some comments and would like to hear your thoughts.

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+  echo "Failed to retrieve current logVerbosity value."
+  exit 1
+fi
+
+echo
+echo "CURRENT LOG VERBOSITY = $current_log_verbosity"
+echo
+
+# Update logVerbosity to 6
+oc patch ca $CA_NAME --type='json' -p='[{"op": "replace", "path": "/spec/logVerbosity", "value": 6}]'
```
bmeng, line 23: The script called the patch operation twice on the resource, and it will trigger the redeploy twice as well.
I feel it should not be the regular task.

scripts/CEE/collect-verbose-ca-logs/metadata.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - "patch"
+        apiGroups:
+        - "autoscaling.openshift.io"
+        resources:
+        - "clusterautoscalers"
+      - verbs:
+        - "get"
+        - "list"
+        apiGroups:
+        - "metrics.k8s.io"
```
bmeng, line 32: Why this api group is needed?

scripts/CEE/collect-verbose-ca-logs/script.sh
feichashao: I assume this is to turn on the verbose to print more info and then turn off the verbose to make it back to normal.
https://access.redhat.com/solutions/6821651
https://access.redhat.com/solutions/6172162
https://docs.openshift.com/container-platform/4.12/machine_management/applying-autoscaling.html#cluster-autoscaler-cr_applying-autoscaling

@kkulshreshth  if that's the case, it would be great if you can share an example output for reviewers' reference.

kkulshreshth: @feichashao You are absolutely right on this. This is to reset the logVerbosity. @bmeng you are right that it will redeploy the pod. Mostly the customers keep logVerbosity to 1 and CA logs with verbosity 1 is literally of no use.

For example, consider the case where a customer has underutilized nodes. With logVerbosity 1, the CA pod prints:
`I0605 09:26:35.309013       1 eligibility.go:102] Scale-down calculation: ignoring 6 nodes unremovable in the last 5m0s
I0605 09:26:35.907516       1 legacy.go:296] No candidates for scale down`

With logVerbosity 6, we get a detailed log highlighting the root cause behind no scale down:
`I0606 07:50:35.265451       1 klogx.go:87] Node osdmgmtdrprd01-c54h8-worker-16-b-q9c4v - cpu utilization 0.417810
I0606 07:50:35.266018       1 cluster.go:155] osdmgmtdrprd01-c54h8-worker-16-b-q9c4v for removal
I0606 07:50:35.266031       1 cluster.go:164] node osdmgmtdrprd01-c54h8-worker-16-b-q9c4v cannot be removed: pod with local storage present: openshift-gitops-repo-server-74694cd9d6-25l52`

**MCS will run this script only after taking customer's consent and only CRU members should be able to run this script.** Additionally, we will integrate this script with our troubleshooting CLI tool (inceptor).

Let me know if you have further questions.

scripts/CEE/collect-verbose-ca-logs/script.sh
kkulshreshth: It is waiting for the pod to redeploy with the new logVerbosity setting. I am checking if I can add a condition to wait for the new CA POD to be in the READY state.

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# Get the name of the Cluster Autoscaler pod
+CA_POD=$(oc get pods -n openshift-machine-api | grep cluster-autoscaler-default | awk '{print $1}')
+
+# Collect logs for the next 6 minutes
+echo
+echo "Collecting logs for the next 6 minutes..."
+echo
+
+sleep 360
```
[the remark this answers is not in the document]
kkulshreshth: In most of the cases, the CA checks the eligibility of a node for scale down if it is unneeded for 5 minutes OR 10 minutes. This is something that the customer can configure. Previously, I thought it was default to 5 but while working on a recent case I realize it could be 10 as well. TBH, I kept it 6 because in most of the case it worked for me. 

However, after some research I got to know about `--scale-down-unneeded-time=5m` arg which is defined in the CA Pod YAML. In my code, I will try to get this value and the script will collect the logs for **"scale-down-unneeded-time + 1" minutes.**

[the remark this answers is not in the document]
kkulshreshth: Sorry, I just checked that the arg is only available if the customer configures it. I believe collecting it for 6  minutes should be fine as if we redeploy the pod, it will immediately start checking all the nodes. This will help us keep the script simple.

scripts/CEE/collect-verbose-ca-logs/metadata.yaml
kkulshreshth: This API group is not needed, I have removed it.

scripts/CEE/collect-verbose-ca-logs/script.sh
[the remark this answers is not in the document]
feichashao: thanks for explaining! that makes sense.

scripts/CEE/collect-verbose-ca-logs/script.sh
kkulshreshth: Check condition added in the script.

scripts/CEE/collect-verbose-ca-logs/script.sh
feichashao: thanks for explaining this

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+
```
feichashao, line 2: Suggest to add
```
set -e
set -o nounset
set -o pipefail
```
eg https://github.com/openshift/managed-scripts/blob/8d6d682428d559461ddd209ca993f86067bd99f5/scripts/CEE/delete-worker-node/script.sh#L3C1-L5C16

This arguments can safeguard the script in case any failures in the middle.

kkulshreshth: I added these arguments and retested the script, it is working as expected. Adding the updated script file.

scripts/CEE/collect-verbose-ca-logs/metadata.yaml
```diff
@@ -0,0 +1,42 @@
...
+        resources:
+        - "clusterautoscalers"
+      - verbs:
+        - "get"
+        - "list"
+        apiGroups:
+        - ""
+        resources:
+        - "nodes"
+      - verbs:
```
bmeng, line 35: duplicated permission

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,124 @@
...
+set -o pipefail
+
+# Set variables
+GREEN="\033[32m"
+RESET="\033[0m"
+NAMESPACE="openshift-machine-api"
+CA_NAME="default"
+
+# Get current logVerbosity value
+current_log_verbosity=$(oc get ca $CA_NAME -o jsonpath='{.spec.logVerbosity}')
```
bmeng, line 14: Can we use the full resource name `clusterautoscalers` here rather than the short version?

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,124 @@
...
+    break
+  fi
+
+  # Print the current status and wait before rechecking
+  echo "Current status: $POD_STATUS. Waiting..."
+  sleep 5
+done
+
+# Get the name of the Cluster Autoscaler pod
+CA_POD=$(oc get pods -n openshift-machine-api | grep cluster-autoscaler-default | awk '{print $1}')
```
bmeng, line 59: Do we need to get it again here?

scripts/CEE/collect-verbose-ca-logs/script.sh
```diff
@@ -0,0 +1,124 @@
...
+echo
+
+# Sleep for 6 minutes to let CA pod generate logs
+sleep 360
+
+echo "---------------------"
+echo "LOG COLLECTION: START"
+echo "---------------------"
+echo
+CA_LOGS=$(oc logs -n $NAMESPACE "$CA_POD" --since=6m)
```
bmeng, line 73: Would it be better if we write this to file instead of set it as variable?

scripts/CEE/collect-verbose-ca-logs/script.sh
bmeng: the same as the other places.

scripts/CEE/collect-verbose-ca-logs/metadata.yaml
kkulshreshth: Copy paste mistake, I have removed the duplicate part.

scripts/CEE/collect-verbose-ca-logs/script.sh
kkulshreshth: Not required as we are already reading it while implementing the check condition for nee pod's running state. I have removed #58 and #59 lines.

scripts/CEE/collect-verbose-ca-logs/script.sh
kkulshreshth: We will further use the output of this script in our CLI tool to suggest next steps and KCS solutions.

PR #189, opened 2024-06-04
osd23682: consolidating cordon/drain/reboot worker nodes

faldanarh:
### What type of PR is this?

Adding a new script for consolidating cordon/drain/reboot worker nodes.

This PR is to add the new script.

_Resolves #_ https://issues.redhat.com/browse/OSD-23682


feichashao:
@faldanarh   thanks for the PR!  The build failed with the shellcheck, please review:
```
09:49:54 In ./CEE/cordon-drain-reboot-worker-node/script.sh line 88:
09:49:54     if [ -z ${1} ]; then
09:49:54             ^-- SC2086: Double quote to prevent globbing and word splitting.
09:49:54 
09:49:54 
09:49:54 In ./CEE/cordon-drain-reboot-worker-node/script.sh line 96:
09:49:54             ${DRAIN_CMD} ${DRAINMODE}
09:49:54                          ^-- SC2086: Double quote to prevent globbing and word splitting.
09:49:54 
09:49:54 
09:49:54 In ./CEE/cordon-drain-reboot-worker-node/script.sh line 101:
09:49:54         ${DRAIN_CMD} ${1}
09:49:54                      ^-- SC2086: Double quote to prevent globbing and word splitting.
09:49:54 
09:49:54 
09:49:54 In ./CEE/cordon-drain-reboot-worker-node/script.sh line 118:
09:49:54     drain_worker ${FORCEDRAINMODE}
09:49:54                  ^-- SC2086: Double quote to prevent globbing and word splitting.
```

scripts/CEE/drain-reboot-worker-node/metadata.yaml
```diff
@@ -31,15 +31,18 @@ rbac:
         - ""
         resources:
         - "nodes"
+        - "nodes/proxy"
```
feichashao, line 34: Could you help me to understand why this file is also changed? It seems to be in a different folder than the other 3 files.

scripts/CEE/cordon-drain-reboot-worker-node/script.sh
```diff
@@ -0,0 +1,186 @@
...
+            ${DRAIN_CMD}
+        else 
+            echo "Draining node ${WORKER} with drain mode '${DRAINMODE}'"
+            echo
+            ${DRAIN_CMD} ${DRAINMODE}
+        fi
+    else
+        echo "Draining node ${WORKER} for rebooting."
+        echo
+        ${DRAIN_CMD} ${1}
```
feichashao, line 101: This may fail shellcheck, maybe we can make the options fixed.
If mode==force then execute xxx
else execute yyy.

bmeng:
Would you like to deprecate the standalone scripts as well?

scripts/CEE/cordon-drain-reboot-worker-node/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Usage
+
+**ATTENTION** ⚠️ This script must only be used by the MCS (Managed Cloud Services) CRU (Crew Response Unit) team.
+- If you need this script to be used, contact a MCS CRU member.
+- The usage of this script is audited.
+
+### Cordon/Uncordon
+
+```bash
+ocm backplane managedjob create CEE/cordon-drain-reboot-worker-node -p WORKER="<node_name>" -p ACTION="[cordon|uncordon]"
```
bmeng, line 25: I am thinking maybe we can give it a more common name, like `worker-node-operations`

scripts/CEE/drain-reboot-worker-node/metadata.yaml
faldanarh: Yep, although this also could apply here, we can disregard this change as it was by mistake.

scripts/CEE/cordon-drain-reboot-worker-node/script.sh
faldanarh: I don't want to make it fixed because drain has a lot of options we can use. So passing it as a variable would helps to choose whatever is needed. I tested and it ran ok.

scripts/CEE/cordon-drain-reboot-worker-node/README.md
faldanarh: Yes, @bmeng , the idea is to deprecate those duplicate.
And about the name, `worker-node-operations` is not bad, but it will not be clear which operations we are referring to, just on the description. But you guys decide which is better.

scripts/CEE/cordon-drain-reboot-worker-node/script.sh
Dee-6777: @faldanarh The shell check is failing here. If don't want to change it, could you please add a double quote to the variables here?

scripts/CEE/cordon-drain-reboot-worker-node/README.md
Dee-6777: I would give a +1 to the suggestion above. The operations are already mentioned in the Readme file.

scripts/CEE/cordon-drain-reboot-worker-node/script.sh
feichashao: Hi @faldanarh , I got your point that we want to make the drain command flexible. The `shellcheck` force you to add a double quote for a reason. With a double quote, it will take the whole string as one single argument, so that it can avoid users hijacking the command. Eg, the user can pass "; rm -rf /;" as arguments.

scripts/CEE/cordon-drain-reboot-worker-node/README.md
bmeng: I was just thinking, if we have more operations to add here, and the script name would be no sense again, and we may need to decide either to create a new script for replacement or create a separated script for that part.

faldanarh: Shall I change it back, @bmeng ?
The name it was does not matter as it will only describe what it is and be copied and pasted.

bmeng:
I am a bit not sure if we should combine them together or just let them be separated.
Putting things together means we need to grant more permissions to the single place, and will also give more options to the script. Means people can do more things with the single script also it is easy to make mistake.

faldanarh:
> I am a bit not sure if we should combine them together or just let them be separated. Putting things together means we need to grant more permissions to the single place, and will also give more options to the script. Means people can do more things with the single script also it is easy to make mistake.

So, maybe it is better to leave how it was then.

PR #190, opened 2024-06-12
[OSD-19695] make sure the binaries are up to date

bmeng:
### What type of PR is this?

_feature_

Make the binaries in the image are up to date

_Resolves #[OSD-19695](https://issues.redhat.com//browse/OSD-19695)_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.9 AS build-stage0
+
 ARG OC_VERSION="stable"
 ENV OC_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OC_VERSION}"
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -90,12 +94,21 @@ WORKDIR /ocm-cli
...
 
 # Checking the SHA-256 hash
-RUN echo $OCM_SHA256_CHECKSUM | sha256sum -c -
+RUN curl -sSLf $OCM_SHA256_CHECKSUM_URL | sha256sum -c -
 
 # Moving the validated ocm binary to /out and make it executable
 RUN mv ocm-linux-amd64 /out/ocm
 RUN chmod +x /out/ocm
 
+# Install golang with specific version
```
xiaoyu74, line 103: I built a docker image based on your updated Dockerfile and run a container
```
docker run -it shawn-test /bin/bash
[root@cdb748e8bb57 ~]#
[root@cdb748e8bb57 ~]# go version
bash: go: command not found

echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
Looks  no `go path` included.  Can you double check? 

The `ocm version` and `yq --version` look all good.

bmeng: The go installed in the builder only, for building the hypershift, and it will copy the hypershift binary only to the final image

xiaoyu74: oh, I see. Thanks. 
I can see the Acceptance Criteria also mentions: `An automated dependency upgrade system for binaries set up in the managed scripts repository.`  If it's an optional and not considering in this PR, then LGTM.

bmeng: As long as all the binaries/packages can be up to date all the time. I think it is kind of covered.

PR #192, opened 2024-06-12
OSD-23925: flag customer data facing scripts

typeid:
### What type of PR is this?

feature

This PR flags user data accessing scripts as such as part of the lockbox enablement. The flag implementation will be done as part of https://issues.redhat.com/browse/OSD-22636 

Resolves https://issues.redhat.com/browse/OSD-23925

### DIFF
--- a/hack/metadata.schema.json
+++ b/hack/metadata.schema.json
@@ -112,7 +112,11 @@
         "type": "string",
         "pattern": "^\\d+\\.\\d+\\.\\d+$|^\\d+\\.\\d+\\.\\*$"
       }
-    }
+    },
+    "customerDataAccess": {
+      "type": "boolean",
+      "description": "Wether or not this script has access to customer data"
+  }
   },
   "additionalProperties": false,
   "required": [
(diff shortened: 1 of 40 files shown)


RaphaelBut:
Nice work @typeid , I think we will need to add the new label to the schema validation with something like 
```
+    "customerDataAccess": {
+      "type": "boolean",
+      "description": "Wether or not this script has access to customer data"
     }
```
and make it a required field. Wdyt?

typeid:
/hold
Waiting on https://issues.redhat.com/browse/OSD-22636

hack/metadata.schema.json
```diff
@@ -112,7 +112,11 @@
...
       }
-    }
+    },
+    "customerDataAccess": {
+      "type": "boolean",
+      "description": "Wether or not this script has access to customer data"
+  }
   },
   "additionalProperties": false,
   "required": [
```
RaphaelBut, line 122: I would say lets make it a required field to have it consistently in every script. We could have it optionally and only fail when its missing and the target cluster is access protected. But I would be strongly in favor of making it required, so its easier to build on top of it later. wdyt?

PR #193, opened 2024-06-12
add :Z mount to fix local validation errors when using podman

RaphaelBut:
not sure if this is the right approach but when calling `make validation ` using podman I get permissions errors.

### DIFF
--- a/hack/schema_validation.sh
+++ b/hack/schema_validation.sh
@@ -61,7 +61,7 @@ yamlList() {
 check_validation() {
 
   echo "validating the jsonschema for $yamlFiles"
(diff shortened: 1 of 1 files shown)


RaphaelBut:
I have not tested that this still works with docker

hack/schema_validation.sh
```diff
@@ -61,7 +61,7 @@ yamlList() {
 check_validation() {
```
Tessg22, line 62: Shouldn't we add error handling to capture potential errors during execution?
```
  # Ensure the function receives the required parameter
  if [ -z "$1" ]; then
    echo "Error: No YAML files specified."
    return 1
  fi

  local yaml_files="$1"
```

Tessg22 (changes requested):

bmeng:
Hi @RaphaelBut any update for this PR? Is it still needed?

xiaoyu74:
@RaphaelBut - Please rebase the PR to resolve the conflicts. Also, any updates for the PR based on the current review feedback?

PR #196, opened 2024-07-17
OSD-24608 : fixed docker file build issue and SElinux container execute issues 

samanthajayasinghe:
### What type of PR is this?

bug

Docker build throws the following errors. 
```
oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by oc)
```

_Resolves #_
https://issues.redhat.com/browse/OSD-24608

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.9 AS build-stage0
 
-ARG OC_VERSION="stable"
+ARG OC_VERSION="stable-4.15"
 ENV OC_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OC_VERSION}"
 
 # AWS cli args and env variables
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.9 AS build-stage0
 
-ARG OC_VERSION="stable"
+ARG OC_VERSION="stable-4.15"
```
MitaliBhalla, line 3: Are we sure binding it to a particular version?

MitaliBhalla (changes requested):

Dockerfile
samanthajayasinghe: seems . latest version of `oc` required `GLIBC_2.33` and ubi8/ubi:8.9 image may not include the latest `GLIBC` version .. 
It's not nice . we need to create a card to fix the old image :(

PR #199, opened 2024-08-02
[OSD-24941] Update make file to use EPEL8 and fix all the shell check errors

Dee-6777:
### What type of PR is this?

_bug_

### What this PR does / Why we need it?
It will update the scripts as the shell check points out new errors. We've switched to EPEL8 as Centos7/RHEL7 reached EOL a month ago and they moved EPEL7 to archive [https://dl.fedoraproject.org/pub/epel/](https://dl.fedoraproject.org/pub/epel/) [https://dl.fedoraproject.org/pub/archive/epel/](https://dl.fedoraproject.org/pub/archive/epel/).

### Which Jira/Github issue(s) does this PR fix?

_Resolves_ #[OSD-24941](https://issues.redhat.com/browse/OSD-24941)

### Pre-checks (if applicable)

- [ ] Validated the changes in a cluster
- [ ] Included documentation changes with PR

PR #200, opened 2024-08-26
Update script.sh for managedjob CEE/collect-verbose-ca-logs

kkulshreshth:
### Description
Modifying the message that gets displayed if the script is unable to retrieve logVebosity value. The descriptive message will help the user to verify if the autoscaling is properly enabled. Additionally, increasing the sleep time from 5 seconds to 10 seconds to address interim issues in pod name discrepancy.

### What type of PR is this?
Script improvement

The PR has following changes:
- Modifying the message that gets displayed if the script is unable to retrieve logVebosity value.
- Increasing the sleep time from 5 seconds to 10 seconds to address interim issues in pod name discrepancy.


feichashao:
Need to wait for https://github.com/openshift/managed-scripts/pull/199 to unblock the shellcheck.

PR #206, opened 2024-12-13
Feat/delete os pod

alvlkov:
### What type of PR is this?

This adds a new managed script to delete a pod from Openshift's reserved namespace.

This will help fixing errors related to openshift reserved namespaces, essentially when pod restart is required.

[OSD_20528](https://issues.redhat.com/browse/OSD-20528)

- [x] Validated the changes in a ROSA stage cluster


scripts/CEE/delete-os-pod/README.md
```diff
@@ -0,0 +1,21 @@
+# Delete Openshift Pod Script
```
iamkirkbater, line 1: Can we simplify this to just be `delete-pod` instead of adding the `delete-os-pod`? From a UX perspective, it will be easier to remember the closer the syntax name is to the actual OC command.

scripts/CEE/delete-os-pod/script.sh
```diff
@@ -0,0 +1,47 @@
...
+  if [ $? -eq 0 ]; then
+    echo -e "\n[SUCCESS] Pod '$POD_NAME' successfully deleted from namespace '$NAMESPACE'."
+  else
+    echo -e "\n[ERROR] Failed to delete pod '$POD_NAME' from namespace '$NAMESPACE'."
+  fi
+}
+
+
+main(){
+  delete_pod
```
iamkirkbater, line 44: Would it be a huge lift here to validate if a pod is owned by a replicaset before proceeding? We might also need to add a "force" flag/parameter to bypass that as well, but it might be a nice protection for the rare chance that a pod isn't managed in an openshift namespace, this way we can make sure it will come back as a default behavior, but have the option to bypass it if we need to.

iamkirkbater (changes requested):
Requested change relates to the name of the script. The additional check for a replicaset would be more of a nice-to-have, but we can also add that _after_ this is merged so that we can start using this sooner rather than later.

scripts/CEE/delete-os-pod/metadata.yaml
```diff
@@ -0,0 +1,28 @@
+file: script.sh
+name: delete-os-pod
+shortDescription: Deletes a pod from openshift namespace 
+description: Deletes a single pod from openshift's reserved namespace. 
+author: Alex Volkov
+allowedGroups:
+  - CEE
+  - SREP
```
iamkirkbater, line 8:
```suggestion
  - SREP
  - MCSTierTwo
```

feichashao:
Thanks @iamkirkbater for the review!

I would suggest we add the safeguard in this PR to validate if a pod is backed by a replicaset, otherwise the delete operation can be too wide.
The protection can be "we are not making the situation worse":
- If we are deleting a non-healthy pod, go ahead and delete it.
- If we are deleting a healthy pod,
    - If it is the only healthy pod in the replicaset, stop, raise a ticket and review it.
    - If there's another healthy pod besides the one we are going to delete, it is ok to delete.

Another nice-to-have is that we put a list of allowed namespace instead of `openshift-*`. This sound like a toil but it give us an opportunity to review if we want to allow the deletion when a new namespace comes. (can be next PR for this one).

feichashao:
> Can you expand on what you mean by "non-healthy" pod? If we're asking for this in this PR I'd like to be explicit to what we are looking for. For example, if we just mean a "healthy" pod is one in a "Running" state, vs `non-healthy` which would be "Error", "Completed", "Pending" - etc.

I would say Healthy = A pod with all containers in running state; The other should be non-healthy, eg, pending, crashloopbackoff, pod in running state but not all containers are running, showing like:
```
kube-apiserver-ip-10-119-135-4.ec2.internal           4/5     Running 
```
(I mocked this)

scripts/CEE/delete-os-pod/metadata.yaml
alvlkov: added the suggestions, thanks @iamkirkbater

alvlkov:
Added replicaset check and --force flag.
- [x] - Successfully deleted pod owned by replicaset regardless --force flag
- [x] - Couldn't delete a pod not owned by a replicaset without --force flag
- [x] - Successfully deleted pod not owned by replicaset with --force flag

scripts/CEE/delete-pod/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+      verbs:
+        - "delete"
+        - "get"
+
+envs:
+- key: POD_NAME
+  description: Name of the pod to delete
+  optional: false
+- key: NAMESPACE
+  description: Namespace name where por to delete is running, must start with openshift-*
```
typeid, line 24:
```suggestion
  description: Namespace name where pod to delete is running, must start with openshift-*
```

scripts/CEE/delete-pod/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - MCSTierTwo
+rbac:
+  clusterRoleRules:
+    - apiGroups:
+        - ""
+      resources:
+        - "pods"
+      verbs:
+        - "delete"
+        - "get"
```
typeid, line 17: This is valid for all namespaces. There's no limitation to `from openshift's reserved namespace.` as mentioned above. 

This permission extends beyond the scope even SRE-P has.

alvlkov: The above limitation applies to the NAMESPACE parameter, to avoid deleting Openshift related pods.  AFAIK I cant scope namespaces within clusterRoleRules. Please elaborate about the suggestion.

typeid:
/unhold

Merging this as we have not received any feedback from compliance. This does not provide read access to customer data so I'm okay just stamping this off.

PR #208, opened 2024-12-17
Feat/check target connectivity

alvlkov:
### What type of PR is this?

This adds a new managed script to perform multiple checks to validate external target connectivity from Openshift cluster. 

[OSD-26447](https://issues.redhat.com/browse/OSD-26447)

### DIFF
--- a/scripts/CEE/check-target-connectivity/metadata.yaml
+++ b/scripts/CEE/check-target-connectivity/metadata.yaml
@@ -1,0 +1,48 @@
+file: script.sh
+name: check-target-connectivity
+shortDescription: Performs multiple checks to validate target connectivity.
+description: |
+  Performs multiple checks to validate target connectivity. 
+
+  Performed checks:
+  - DNS resolution check via nslookup: $ nslookup "$TARGET"
+  - DNS resolution via Dig: $ dig +short "$TARGET"
+  - ICMP check via ping: $ timeout 10 ping -c 3 "$TARGET"
+  - Routing Check via traceroute: $ timeout 5 traceroute -m 10 -w 1 -q 1 "$TARGET"
+  - Check Target Port is Open via nmap: $ timeout 5 nmap -p "$PORT" "$TARGET" 2>&1 | grep -q "$PORT/tcp open"
+
+  Notes: 
+  - Each check awaits for 5 seconds before starting to minimize impact on the network. 
+
+author: Alex Volkov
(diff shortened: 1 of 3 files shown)


feichashao:
Hi @alvlkov , thank you for raising this PR and thank you for including a README!

The managed-script itself will run as pod in the cluster. I see the script is calling `oc` create another pod and runs the network utils in the new pod, is there a specific reason to run them in a new pod?

alvlkov:
> Hi @alvlkov , thank you for raising this PR and thank you for including a README!
> 
> The managed-script itself will run as pod in the cluster. I see the script is calling `oc` create another pod and runs the network utils in the new pod, is there a specific reason to run them in a new pod?

@feichashao thank you for the review!

Yes, there is a particular reason to create another pod. I believe the base image used by managed-script does not provide all necessary dependencies to perform the checks, i.e. nmap:

  - DNS resolution check via nslookup: $ nslookup "$TARGET"
  - DNS resolution via Dig: $ dig +short "$TARGET"
  - ICMP check via ping: $ timeout 10 ping -c 3 "$TARGET"
  - Routing Check via traceroute: $ timeout 5 traceroute -m 10 -w 1 -q 1 "$TARGET"
  - Check Target Port is Open via nmap: $ timeout 5 nmap -p "$PORT" "$TARGET" 2>&1 | grep -q "$PORT/tcp open"
  
  Other scripts as PCAP collector use the same approach and image:  quay.io/app-sre/srep-network-toolbox:latest is suitable.

feichashao:
@alvlkov thanks for the clarifications! That make sense if the image itself doesn't have the needed binaries. Probably we can add it to the image https://github.com/openshift/managed-scripts/blob/main/Dockerfile but let me discuss with the team.

scripts/CEE/check-target-connectivity/metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - "pods/exec"
+      - "pods/log"
+    verbs:
+      - "create"
+      - "get"
+      - "delete"
+  - apiGroups:
+      - "security.openshift.io"
+    verbs:
+      - "*"
```
bmeng, line 36: why do we need all permissions to SCC?

alvlkov: changed to:
    verbs:
      - "use"

PR #210, opened 2025-02-06
Add validate pull-secret script to the SREP scripts

zmird-r:
https://issues.redhat.com/browse/OSD-27688

### DIFF
--- a/scripts/SREP/validate-pull-secret/script.sh
+++ b/scripts/SREP/validate-pull-secret/script.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+
+DEBUG=0
+
+# define usage
+usage() {
+  cat <<EOF
+usage: validate-pull-secret -d
+
+optional flags:
+  -d | -debug        - Enable debug output
+EOF
+}
+
+function print_debug() {
+  if [[ $DEBUG -ne 0 ]]; then
+    echo "[DEBUG] $1"
+  fi
+}
+
+while [ "$1" != "" ]; do
+  case $1 in
+    -d | --debug )          shift
+                            DEBUG=1
+                            ;;
+    * ) echo "Unexpected parameter $1"
+        usage
+        exit 1
+  esac
+  shift
+done
+
+# Retrieve UUID
+CLUSTER_UUID=$(oc get clusterversion version -o json | jq -r '.spec.clusterID')
+
+# Retrieve ID
+CLUSTER_ID=$(ocm list clusters --parameter search="external_id is '${CLUSTER_UUID}'" --columns id --padding 20 --no-headers)
+
+print_debug "Retrieved cluster UUID $CLUSTER_UUID ID $CLUSTER_ID"
+
+# Retrieve user
+SUBSCRIPTION_ID=$(ocm get /api/clusters_mgmt/v1/"$CLUSTER_ID" | jq -r '.subscription.id')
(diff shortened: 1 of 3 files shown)


scripts/SREP/validate-pull-secret/script.sh
```diff
@@ -0,0 +1,81 @@
...
+        exit 1
+  esac
+  shift
+done
+
+# Retrieve UUID
+CLUSTER_UUID=$(oc get clusterversion version -o json | jq -r '.spec.clusterID')
+
+# Retrieve ID
+CLUSTER_ID=$(ocm list clusters --parameter search="external_id is '${CLUSTER_UUID}'" --columns id --padding 20 --no-headers)
```
bmeng, line 37: Is it possible to run the `ocm` command inside the script? The script will be executed as job on the target cluster.
Maybe you can pass the cluster id as the parameter?

scripts/SREP/validate-pull-secret/script.sh
```diff
@@ -0,0 +1,81 @@
...
+
+print_debug "Retrieved cluster UUID $CLUSTER_UUID ID $CLUSTER_ID"
+
+# Retrieve user
```
bmeng, line 41: Hmm, seems we have more queries to OCM here.
I think maybe it is not a good use case for managed-scripts.

bmeng:
https://github.com/openshift/managed-scripts/tree/main/scripts/SREP/get-pull-secret-email
we have the existing script to fetch the secret from cluster and integrate with osdctl to do the comparison.

cc @feichashao can you help confirm if this is done?

scripts/SREP/validate-pull-secret/script.sh
zmird-r: I think it's possible, there's other scripts that make use of the same command https://github.com/openshift/managed-scripts/blob/master/scripts/SREP/update-upgrade-policy-state/script.sh#L29

What would be the advantage of having the cluster ID passed as a parameter? Looking at other scripts they don't require the cluster ID e.g. https://github.com/openshift/managed-scripts/blob/main/scripts/SREP/describe-nodes/README.md

scripts/SREP/validate-pull-secret/script.sh
zmird-r: I've addressed the OCM concern in the other comment, can you expand on why it would not be a good use case?

bmeng: ok, the two comments are the same point, let's use this for discussion. I am not aware of any other script uses ocm as well. I doubt even for that one it may not be working well. 
As I mentioned, the managed-script will create a kubernetes job which is running in the cluster. I cannot understand how a ocm call can be succeeded without login.
Generally we should execute the cluster related (means kubernetes or openshift here) commands via the managed-scripts, but not the ocm.

Have you tested the script by yourself to see if it is working? (I can have a test next week as well.) 

Please ignore the other comment where I suggest to pass the cluster id as parameter, that is not the right use case as well.

Dee-6777: @zmird-r you can extract the internal cluster id from configmap through this command:
```
oc get configmap cluster-config-v1 -n kube-system -o jsonpath='{.data.install-config}' | grep -oE 'api.openshift.com/id: [^ ]+' | awk '{print $2}'
```

Dee-6777: Can you try using this instead of `ocm`?

zmird-r:
Without being able to use the ocm cli this script is not possible at the moment. Closing the ticket

PR #230, opened 2025-08-26
Add osdctl to dockerfile

dakotalongRH:
### What type of PR is this?

Add osdctl to Docker image for managed scripts

https://issues.redhat.com/browse/SREP-861
_Resolves #_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,14 +119,25 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
 # Build binary
 RUN OUT_DIR=/out make hypershift
 
-# Attach osdctl binary into the image
-# Fetch the source code
+## Attach osdctl binary into the image
+# Setting ENV variables
+ENV OSDCTL_URL="https://github.com/openshift/osdctl/releases/latest/download/osdctl_0.46.0_Linux_x86_64.tar.gz"
+ENV OSDCTL_SHA256_CHECKSUM_URL="https://github.com/openshift/osdctl/releases/latest/download/sha256sum.txt"
+
+# Creating a working directory
 RUN mkdir -p /osdctl
 WORKDIR /osdctl
-RUN git clone https://github.com/openshift/osdctl.git /osdctl
-# Build binary (ci-build will download goreleaser first, then build only for current platform)
-RUN SINGLE_TARGET=true make ci-build
-RUN cp dist/osdctl_linux_amd64_v1/osdctl /out/osdctl
+
+# Downloading the osdctl binary
+RUN curl -sSLf -O $OSDCTL_URL
+
+# Checking the SHA-256 hash
+RUN curl -sSLf $OSDCTL_SHA256_CHECKSUM_URL | grep "osdctl_0.46.0_Linux_x86_64.tar.gz" | sha256sum -c -
+
+# Extract the tarball and move the validated osdctl binary to /out
+RUN tar -xzf osdctl_0.46.0_Linux_x86_64.tar.gz
+RUN mv osdctl /out/osdctl
+RUN chmod +x /out/osdctl
 
 # Make binaries executable
 RUN chmod -R +x /out
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,8 +124,8 @@ RUN OUT_DIR=/out make hypershift
 RUN mkdir -p /osdctl
 WORKDIR /osdctl
 RUN git clone https://github.com/openshift/osdctl.git /osdctl
-# Build binary (ci-build will download goreleaser first, then build)
-RUN make ci-build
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ ARG AWSCLI_VERSION="awscli-exe-linux-x86_64.zip"
 ENV AWSCLI_URL="https://awscli.amazonaws.com/${AWSCLI_VERSION}"
 
 # install tools needed for installation
-RUN yum install -y unzip git make gcc
+RUN dnf install -y unzip git make gcc
```
joshbranham, line 13: What's the reasoning for changing from `yum` to `dnf` across the Dockerfile?

Dockerfile
```diff
@@ -119,20 +119,30 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
 # Build binary
 RUN OUT_DIR=/out make hypershift
 
+# Attach osdctl binary into the image
+# Fetch the source code
+RUN mkdir -p /osdctl
+WORKDIR /osdctl
+RUN git clone https://github.com/openshift/osdctl.git /osdctl
+# Build binary (ci-build will download goreleaser first, then build only for current platform)
+RUN SINGLE_TARGET=true make ci-build
```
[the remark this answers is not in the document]
dakotalongRH: Just tried this instead and it's actually a fair bit faster, I'll update to use this method instead

Dockerfile
joshbranham: We discussed offline that they are essentially the same, but I suggested making the change in a standalone PR.

Dockerfile
```diff
@@ -119,20 +119,41 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
 # Build binary
 RUN OUT_DIR=/out make hypershift
 
+## Attach osdctl binary into the image
+# Setting ENV variables
+ENV OSDCTL_URL="https://github.com/openshift/osdctl/releases/latest/download/osdctl_0.46.0_Linux_x86_64.tar.gz"
+ENV OSDCTL_SHA256_CHECKSUM_URL="https://github.com/openshift/osdctl/releases/latest/download/sha256sum.txt"
```
joshbranham, line 125: This should match the version we are downloading, otherwise when we push the next version, the checksum will fail.

Dockerfile
```diff
@@ -119,20 +119,41 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
...
+
+# Creating a working directory
+RUN mkdir -p /osdctl
+WORKDIR /osdctl
+
+# Downloading the osdctl binary
+RUN curl -sSLf -O $OSDCTL_URL
+
+# Checking the SHA-256 hash
+RUN curl -sSLf $OSDCTL_SHA256_CHECKSUM_URL | grep "osdctl_0.46.0_Linux_x86_64.tar.gz" | sha256sum -c -
```
joshbranham, line 135: If you change the version you have to change this line too, factor that out a bit so you only set the version in one place.

Dockerfile
```diff
@@ -119,20 +119,41 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
...
+
+# Extract the tarball and move the validated osdctl binary to /out
+RUN tar -xzf osdctl_0.46.0_Linux_x86_64.tar.gz
```
joshbranham, line 138: Same comment about version in one place.

Dockerfile
```diff
@@ -119,20 +119,41 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
...
+RUN mv osdctl /out/osdctl
+RUN chmod +x /out/osdctl
+
 # Make binaries executable
 RUN chmod -R +x /out
 
 FROM registry.access.redhat.com/ubi8/ubi:8.9
-RUN  yum -y install --disableplugin=subscription-manager \
```
joshbranham, line 126: Do you need to remove this flag in your PR?

Dockerfile
```diff
@@ -119,20 +119,41 @@ RUN git clone https://github.com/openshift/hypershift.git /hypershift
...
+RUN  yum -y install \
      python3.11 python3.11-pip jq openssh-clients sshpass \
-     && yum --disableplugin=subscription-manager clean all
```
joshbranham, line 128: Same comment as above about the flag.

Dockerfile
dakotalongRH: This is from testing where I faced the issue with subscription manager on a Fedora machine (and a subsequent quick test in CI), it does not need to be in the actual PR

Dockerfile
dakotalongRH: version is grabbed dynamically now

PR #250, opened 2026-03-27
Fix konflux PR check

RaphaelBut:
[Fix IntegrationTestScenario PR check pipeline for Konflux](https://github.com/openshift/managed-scripts/pull/250/changes/b253f3113689a94a0ffd0a1ffa8eb5c67cb91d78)
Changes to make the PR validation work with Konflux IntegrationTestScenario:

- Removed duplicate SNAPSHOT parameter (Konflux injects it automatically)
- Changed from taskRef to inline taskSpec (task-bash bundle doesn't exist)
- Run validation tools directly to avoid nested container issues
- Use python3.11 -m check_jsonschema instead of command-line tool
- Set correct workingDir to $(workspaces.source.path)/source

This allows the PR check to run metadata validation, shellcheck, and
pyflakes without requiring nested containers or external task bundles.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/.tekton/pr-check-pipeline.yaml
+++ b/.tekton/pr-check-pipeline.yaml
@@ -6,9 +6,6 @@ metadata:
     pipelinesascode.tekton.dev/on-event: "[pull_request]"
     pipelinesascode.tekton.dev/on-target-branch: "[main]"
 spec:
-  params:
-    - name: SNAPSHOT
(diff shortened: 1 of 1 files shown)


MateSaary (approved):
/lgtm

PR #252, opened 2026-03-27
Remove duplicate test exection caused by pipelinesascode

RaphaelBut:
The PR check pipeline is trigger by a integrationtestscenario in konflux. 
These annotations caused it to be executed twice.

### What type of PR is this?
bug


MateSaary (approved):
/lgtm

PR #255, opened 2026-03-30
Fix PR check to run against pr revision

RaphaelBut:
### What type of PR is this?

bug

Pr check was always running against main instead of the revision of the PR

https://redhat.atlassian.net/browse/SREP-2829

_Resolves #_


MateSaary (approved):
/lgtm

